PR #2, opened 2024-09-12
Templater - generating ISPyPSA templates from Assumptions workbook

prakaa:
Produces templates with merged in values for:
- Flow paths (lines)
  - Regional or sub-regional 
- Nodes (buses)
  - Regional or sub-regional
- Existing, committed, anticipated and additional (projects) generators
  - Static (i.e. not time-varying) values merged in
  - This model will be used for new entrants and storage

Cached workbook tables and templates are produced by a doit pipeline. Run `doit` in the top directory and this will initiate the pipeline


prakaa:
@nick-gorman @dylanjmcconnel any thoughts on the following:

- Repo structure 
  - Just like `isp-workbook-parser`, mappings packaged within `src`
  - A `templater` submodule (is this the best name? it turns it into template format), with a submodule that assembles components from assumptions workbooks tables
  - Exposed function in `components.py` that assembles and references a private function that does the actual processing
- Templater mapping
  - I have a version level in there, but this may be overkill?

prakaa:
After talking to Nick, I think I got confused about what the templater should do. It shouldn't use PyPSA naming convention (i.e. we have a translator to create PyPSA-friendly output).

It should:

- Use AEMO naming convention where possible, but column names are in snakecase
- Only wrangle the table if it improves the editability of the table for the user (e.g. sub-regional network)

prakaa:
@nick-gorman this is ready to review

PR #4, opened 2024-10-28
Add logging config and add logging to templater and fuzzy matching

### LINKED ISSUE #3: fuzzy matching and logging
I think we should implement logging that records any fuzzy matching that results in a non-exact match. This would make it very easy to manually review that the fuzzy matching is functioning as intended and not introducing errors / will help us trust the fuzzy matching overtime as workbook inputs get updated.

prakaa:
- Uses a `configure_logging()` function in `logging.py` to:
  - Disable/configure a logging level for the console/terminal
  - Disable/configure a logging level for the log file
  - Default warning for console, info for file
- Info logging for templater functions
- Info logging functionality added to fuzzy matching:
  - Will give name of original and match
  - Uses a custom task description
    - fuzzy matching is not just used for merging tables in

Example output in ISPyPSA.log:
image
  
Closes #3

### DIFF
--- a/dodo.py
+++ b/dodo.py
@@ -1,14 +1,18 @@
 from pathlib import Path
+
 import yaml
 
 from ispypsa.data_fetch.local_cache import REQUIRED_TABLES, build_local_cache
-from ispypsa.templater.nodes import template_nodes
+from ispypsa.logging import configure_logging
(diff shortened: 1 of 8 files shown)


prakaa:
And for regional granularity: 
image

PR #14, opened 2024-11-12
change config loading and validation to use pydantic

nick-gorman:


nick-gorman:
Currently, I've kept scenario validation as per what was implemented on main, but I'm unsure if the scenario should be validated as the user may create their own model config that does not align with ISP scenarios. Maybe the scenario in the config should actually be called isp_base_scenario or isp_template_scenario?

prakaa:
> Currently, I've kept scenario validation as per what was implemented on main, but I'm unsure if the scenario should be validated as the user may create their own model config that does not align with ISP scenarios. Maybe the scenario in the config should actually be called isp_base_scenario or isp_template_scenario?

Yeah good idea, base scenario could be good. Will change once translator changes are merged in

PR #17, opened 2024-11-25
Refactor network granularity and some other minor changes

prakaa:
## Config
- Restructured: `network` -> `nodes` -> `regional_granularity` 

## Node template
-  now has a type column, and all mappings removed
- option to include rezs as nodes
  
## Mappings
  - `regional_sub_regional` renamed and always saved
  -  new mapping that maps each NEM region to a single sub region always saved
  - released as `mappings_*.csv`

## Dodo
  - workflow folders renamed

## Bus demand timeseries translator
  - Changes to accommodate config changes, but also refactored based on new mappings and to also handle single region


nick-gorman (approved):
I'm happy with these changes.

PR #22, opened 2024-11-27
Logging and output management updates

prakaa:
# Logging
- PyPSA logger propagates base logger (i.e. ISPyPSA) settings

# Config
- Add `ispypsa_run_name`, which is used to specify an output folder containing the various output sub-folders such as `ispypsa_inputs`, `pypsa_friendly` and `outputs`
- Add `path_to_parsed_traces` to `traces` along with validation that checks the path is a folder and that it contains three sub-folders: `demand`, `solar` and `wind`

# Dodo.py
- Propagation of above changes
  - Now you need to specify a config path for a particular model run, and dodo will check deps etc for that model run 
- Cleaning outputs
  - if any of the "intermediate input"  or output generation tasks (i.e. everything except workbook table caching) is rerun, the task will first check if the output folder exists. 
    - If it doesn't, it will create it. 
    - If it does, it will delete the folder ensuring that intermediate output folders reflect the latest output

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -167,10 +167,9 @@ scratch.py
 .~lock*.csv#
 
 # ispypsa ignores
-## CSVs in model_inputs
-model_data/**/*.csv
-model_data/**/*.parquet
-model_data/**/*.hdf5
+ispypsa_runs/**/*.csv
+ispypsa_runs/**/*.parquet
+ispypsa_runs/**/*.hdf5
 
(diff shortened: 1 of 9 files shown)


nick-gorman (approved):
Looks good. I've just added an option to set path_to_parsed_traces to ENV, in which case the path is retrieved from the the env variable PATH_TO_PARSED_TRACES.

PR #23, opened 2024-11-27
Plotting

prakaa:
Work in progress. Starting to build plot functionality.

To date, a flow plot map has been finished, see attachment below
map_of_total_generation_by_carrier


prakaa:
Could add testing for plotting (https://matplotlib.org/stable/api/testing_api.html), but see this discussion: https://stackoverflow.com/questions/27948126/how-can-i-write-unit-tests-against-code-that-uses-matplotlib

PR #24, opened 2024-11-28
refactor snapshot creation and check trace data against snapshot

nick-gorman:
- snapshot definition is moved to the translator
- conversion to 'ns' datetime units for traces is moved to the translator
- functionality added to check traces against snapshot before they are saved by the translator
- Also changes config structure so all config inputs related to temporality sit under the 'temporal' key

### DIFF
--- a/src/ispypsa/translator/snapshot.py
+++ b/src/ispypsa/translator/snapshot.py
@@ -21,11 +21,24 @@ def create_snapshot_index(
         pd.DataFrame
     """
     if year_type == "fy":
-        start_date = datetime(year=start_year - 1, month=7, day=1, hour=0, minute=30)
-        end_date = datetime(year=end_year, month=7, day=1, hour=0, minute=0)
+        start_year = start_year - 1
+        end_year = end_year
+        month = 7
     else:
-        start_date = datetime(year=start_year, month=1, day=1, hour=0, minute=30)
-        end_date = datetime(year=end_year + 1, month=1, day=1, hour=0, minute=0)
+        start_year = start_year
+        end_year = end_year + 1
+        month = 1
+
+    if operational_temporal_resolution_min < 60:
+        hour = 0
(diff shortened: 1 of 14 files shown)


src/ispypsa/translator/snapshot.py
```diff
@@ -0,0 +1,53 @@
+from datetime import datetime
+
+import pandas as pd
+
+
+def create_snapshot_index(
```
prakaa, line 6: Should this be renamed to something like `create_complete_snapshot_index` or `create_full_snapshot_index` to indicate this function is creating a full, pre-aggregation set of snapshots (i.e. there might be a similar function to conduct checks post aggregation)?

src/ispypsa/translator/snapshot.py
```diff
@@ -0,0 +1,53 @@
...
+    """Creates a DatetimeIndex, stored in DataFrame, defining the snapshots for the model before temporal aggregation.
+
+    The index will start at the beginning of `start_year` and finish at the end of
+    `end_year` with the specified temporal resolution.
+
+    Args:
+        start_year: int specifying the start year
+        end_year: int specifying the end year
+        operational_temporal_resolution_min: int specifying the snapshot temporal resolution in minutes
+        year_type: str specifying the year type 'fy' for financial year means that start_year and end_year refer to
```
prakaa, line 21: Full stop after "year type".

prakaa (approved):
Look good @nick-gorman, minor renaming suggestions only

src/ispypsa/translator/snapshot.py
nick-gorman: Good suggestion. I've changed the name to create_complete_snapshot_index.

src/ispypsa/translator/snapshot.py
nick-gorman: Changed.

PR #25, opened 2024-12-01
migrate rye to uv, add isort and add linopy solvers

EllieKallmier:
Added linopy solvers as an optional dependency for the project. 

Also excluded `uv.lock` file from `check-added-large-files` pre-commit hook as the file was 99kB over the 500kB default limit set by this hook.


EllieKallmier:
Updated the min. required version of Python to >=3.10 to fix an issue with Mac silicon and tables<3.10.1, (tables release notes for context [here](http://www.pytables.org/release-notes/RELEASE_NOTES_v3.10.x.html)).

prakaa (approved):

PR #30, opened 2024-12-10
Representative weeks

nick-gorman:
Adds option to filter model snapshot by a list of weeks for each year.

Branched off #24, which should be merged into the main first.

### DIFF
--- a/ispypsa_runs/development/ispypsa_inputs/ispypsa_config.yaml
+++ b/ispypsa_runs/development/ispypsa_inputs/ispypsa_config.yaml
@@ -32,13 +32,14 @@ temporal:
   start_year: 2025
   end_year: 2025
   reference_year_cycle: [2018]
-  # Representative weeks to use instead of full yearly temporal representation.
-  # Options:
-  #   "None": Full yearly temporal representation is used.
-  #   list[int]: a list of integers specifying weeks of year to use as representative. Weeks of year are defined as
-  #   full weeks (Monday-Sunday) falling within the year. For example, if the list is "[0]" the model will only use the
-  #   first full week of each modelled year.
-  representative_weeks: [0, 13, 26, 39]
+  aggregation:
+    # Representative weeks to use instead of full yearly temporal representation.
+    # Options:
+    #   "None": Full yearly temporal representation is used.
+    #   list[int]: a list of integers specifying weeks of year to use as representative. Weeks of year are defined as
+    #   full weeks (Monday-Sunday) falling within the year. For example, if the list is "[1]" the model will only use the
+    #   first full week of each modelled year.
+    representative_weeks: [1, 12, 25, 38]
 # External solver to use
 # Options (refer to https://pypsa.readthedocs.io/en/latest/getting-started/installation.html):
 #   Free, and by default, installed with ISPyPSA:
--- a/ispypsa_runs/development/ispypsa_inputs/ispypsa_config.yaml
+++ b/ispypsa_runs/development/ispypsa_inputs/ispypsa_config.yaml
@@ -32,6 +32,13 @@ temporal:
   start_year: 2025
   end_year: 2025
   reference_year_cycle: [2018]
+  # Representative weeks to use instead of full yearly temporal representation.
+  # Options:
+  #   "None": Full yearly temporal representation is used.
+  #   list[int]: a list of integers specifying weeks of year to use as representative. Weeks of year are defined as
+  #   full weeks (Monday-Sunday) falling within the year. For example, if the list is "[0]" the model will only use the
+  #   first full week of each modelled year.
+  representative_weeks: [0, 13, 26, 39]
 # External solver to use
 # Options (refer to https://pypsa.readthedocs.io/en/latest/getting-started/installation.html):
 #   Free, and by default, installed with ISPyPSA:
--- a/src/ispypsa/config/validators.py
+++ b/src/ispypsa/config/validators.py
@@ -23,6 +23,7 @@ class TemporalConfig(BaseModel):
     start_year: int
     end_year: int
     reference_year_cycle: list[int]
+    representative_weeks: list[int] | None
 
     @field_validator("operational_temporal_resolution_min")
     @classmethod
--- a/src/ispypsa/config/validators.py
+++ b/src/ispypsa/config/validators.py
@@ -16,6 +16,10 @@ class NetworkConfig(BaseModel):
     nodes: NodesConfig
 
 
+class TemporalAggregationConfig(BaseModel):
+    representative_weeks: list[int] | None
+
+
 class TemporalConfig(BaseModel):
     operational_temporal_resolution_min: int
     path_to_parsed_traces: str
@@ -23,7 +27,7 @@ class TemporalConfig(BaseModel):
     start_year: int
     end_year: int
     reference_year_cycle: list[int]
-    representative_weeks: list[int] | None
+    aggregation: TemporalAggregationConfig
 
     @field_validator("operational_temporal_resolution_min")
     @classmethod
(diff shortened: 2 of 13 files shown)


ispypsa_runs/development/ispypsa_inputs/ispypsa_config.yaml
```diff
@@ -24,14 +23,22 @@ network:
...
   start_year: 2025
   end_year: 2025
   reference_year_cycle: [2018]
+  # Representative weeks to use instead of full yearly temporal representation.
+  # Options:
+  #   "None": Full yearly temporal representation is used.
+  #   list[int]: a list of integers specifying weeks of year to use as representative. Weeks of year are defined as
+  #   full weeks (Monday-Sunday) falling within the year. For example, if the list is "[0]" the model will only use the
+  #   first full week of each modelled year.
+  representative_weeks: [0, 13, 26, 39]
```
prakaa, line 41: Could be done now or later, but I think this should be generalised since there may be many aggregation options. I think it could be organised as follows:

1. Create a field under `temporal` called `aggregation`
2. Under aggregation, you can then "select" an option and then provide arguments

Basically, keep as is but add under an aggregation subheading. This then allows different options/input data structures for various aggregation methods

So for example:
```yaml
temporal
...
  aggregation:
    representative_weeks: [0, 13, 26, 39]
```

src/ispypsa/config/validators.py
```diff
@@ -16,12 +16,32 @@ class NetworkConfig(BaseModel):
...
 
-class TraceConfig(BaseModel):
+class TemporalConfig(BaseModel):
+    operational_temporal_resolution_min: int
     path_to_parsed_traces: str
     year_type: Literal["fy", "calendar"]
     start_year: int
     end_year: int
     reference_year_cycle: list[int]
+    representative_weeks: list[int] | None
```
prakaa, line 26: See comment about config organisation. Could create an `AggregationConfig` class with optional fields that are concrete types or refer to validation classes for each type of aggregation

src/ispypsa/translator/temporal_filters.py
```diff
@@ -0,0 +1,182 @@
+from datetime import datetime, timedelta
+
+import pandas as pd
+
+from ispypsa.config.validators import ModelConfig
+
+
+def apply_snapshot_filter(time_series_data, snapshot):
```
prakaa, line 8: Arguments missing type.

 I think the use of `snapshot` in this module and in `snapshot.py` is a bit confusing since snapshots in ISPyPSA can refer to a complete time series (as produced by snapshot.py) or what is attached to a network in PyPSA (which is filtered as per buses.py and generators.py).

Few options:
1. As per the refactor PR, rename the func in snapshot.py to reflect that it is create a full range snapshot or time series
2. The, in this file, you could keep the function name as is. Or, change the function name to `time_series_filter` and change `snapshot` to `snapshot_filter` or `retained_datetimes` or something like that

src/ispypsa/translator/snapshot.py
```diff
@@ -0,0 +1,53 @@
+from datetime import datetime
+
+import pandas as pd
+
+
+def create_snapshot_index(
```
prakaa, line 6: See other PR comment about name and comment in temporal_filters.py

src/ispypsa/translator/temporal_filters.py
```diff
@@ -0,0 +1,182 @@
...
+    12 2020-01-01 12:00:00     12
+    24 2020-01-02 00:00:00     24
+    36 2020-01-02 12:00:00     36
+    48 2020-01-03 00:00:00     48
+
+    """
+    return time_series_data[time_series_data["Datetime"].isin(snapshot.index)]
+
+
+def filter_snapshot(config: ModelConfig, snapshot: pd.DataFrame):
```
prakaa, line 29: Could feed in AggregationConfig if you change structure as suggested in config comment

src/ispypsa/translator/temporal_filters.py
```diff
@@ -0,0 +1,182 @@
...
+            start_year=config.temporal.start_year,
+            end_year=config.temporal.end_year,
+            year_type=config.temporal.year_type,
+        )
+    return snapshot
+
+
+def filter_snapshot_for_representative_weeks(
+    representative_weeks: list[int],
+    snapshot: pd.DataFrame,
```
prakaa, line 87: As per other comments about naming, is the snapshot the "filtered" timeseries fed to PyPSA or the complete time series that is filtered before being fed into PyPSA? I think this arg name should be changed

src/ispypsa/translator/temporal_filters.py
```diff
@@ -0,0 +1,182 @@
...
+        week is 0).
+    """
+    if year_type == "fy":
+        start_year = start_year - 1
+        end_year = end_year
+        month = 7
+    else:
+        start_year = start_year
+        end_year = end_year + 1
+        month = 1
```
prakaa, line 147: Also features in snapshot.py. Could go into translator/helpers.py?

ispypsa_runs/development/ispypsa_inputs/ispypsa_config.yaml
prakaa: Also, while it's nice to be consistent with Python, should we use 1-indexing for things like the first week rather than 0 indexing?

tests/test_temporal_resolution_reduction/test_time_series_representative_weeks_filter.py
```diff
@@ -0,0 +1,107 @@
...
+def test_create_representative_weeks_filter_fail_with_out_of_range_week_number():
+    snapshot = create_snapshot_index(
+        start_year=2024,
+        end_year=2026,
+        operational_temporal_resolution_min=30,
+        year_type="calendar",
+    )
+    with pytest.raises(ValueError):
+        filter_snapshot_for_representative_weeks(
+            representative_weeks=[0, 2, 51],
```
prakaa, line 102: See comment about numbering representative weeks - Python indexing is less interpretable given 52 weeks in the year

prakaa (commented):
See comments around configs structuring, representative weeks indexing and naming of funcs/args

src/ispypsa/translator/snapshot.py
nick-gorman: changed on other PR

ispypsa_runs/development/ispypsa_inputs/ispypsa_config.yaml
nick-gorman: Good suggestions. Made both these changes.

src/ispypsa/config/validators.py
nick-gorman: Made this change.

src/ispypsa/translator/temporal_filters.py
nick-gorman: Good suggestion. Made this change.

src/ispypsa/translator/temporal_filters.py
nick-gorman: Made this change.

src/ispypsa/translator/temporal_filters.py
nick-gorman: I think it OK to leave as is.

tests/test_temporal_resolution_reduction/test_time_series_representative_weeks_filter.py
nick-gorman: Yep, made this change.

src/ispypsa/translator/temporal_filters.py
nick-gorman: Yep, good idea, made the change.

PR #35, opened 2025-01-15
Template new entrant static properties

EllieKallmier:
Added `static_new_generator_properties` to the `templater` module to create `new_entrant_generators.csv` from relevant tables.
- Follows same structure and core functionality as `static_ecaa_generator_properties`, tailored in some instances to fit the new entrants summary tables and data tables. 
- Added tests to check output for data types and expected values, and added necessary tables to test workbook table cache


nick-gorman (commented):
Hi Ellie, looking good!

I've got a few questions and comments, let me know what you think :)

src/ispypsa/templater/static_new_generator_properties.py
```diff
@@ -0,0 +1,325 @@
...
+    df: pd.DataFrame, col: str, csv_data: pd.DataFrame, csv_attrs: dict
+) -> tuple[pd.DataFrame, str]:
+    """Replace values in the provided column of the summary mapping with those
+    in the CSV data using the provided attributes in
+    `_NEW_GENERATOR_STATIC_PROPERTY_TABLE_MAP`
+    """
+    if re.search(r"[vf]om", col):
+        # change mapping from regional cost zone to generator
+        df[col] = df["generator"]
+        # fill BOTN - Cethana row with pd.NA where currently str:
```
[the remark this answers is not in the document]
EllieKallmier: Fixed from v2.4.1 of iso-workbook-parser now :)

src/ispypsa/templater/mappings.py
```diff
@@ -143,3 +144,107 @@
...
+    ),
+    "maintenance_duration_%": dict(
+        csv="maintenance_new_entrants",
+        csv_lookup="Generator type",
+        csv_value="Proportion of time out (%)",
+    ),
+    "fom_$/kw/annum": dict(
+        csv="fixed_opex_new_entrants",
+        csv_lookup="Generator",
+        csv_value="Fixed OPEX ($/kW sent out/year)_NSW Low",
```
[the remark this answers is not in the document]
EllieKallmier: Yep I agree, good note :) Added new separate functions to handle fixed and variable opex merging in in new commit.

src/ispypsa/templater/static_new_generator_properties.py
```diff
@@ -0,0 +1,325 @@
...
+        ]
+    # fills renewable generators, storage and OCGT with 0.0
+    df.loc[
+        _where_any_substring_appears(
+            df[min_level_col], ["solar", "wind", "pumped hydro", "battery", "ocgt"]
+        ),
+        min_level_col,
+    ] = 0.0
+    # replace any remaining cells containing str (tech type) with pd.NA
+    df[min_level_col] = df[min_level_col].apply(
```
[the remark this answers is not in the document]
EllieKallmier: Set to 0.0 for the moment - flagged to come back to as a question issue re: handling this and OCGT min stable levels.

src/ispypsa/templater/static_new_generator_properties.py
```diff
@@ -0,0 +1,421 @@
...
+        threshold=90,
+    )
+    df[col] = df[col].replace(replacement_dict)
+    if "new_col_name" in csv_attrs.keys():
+        df = df.rename(columns={col: csv_attrs["new_col_name"]})
+        col = csv_attrs["new_col_name"]
+    return df, col
+
+
+def _process_and_merge_fixed_opex_base_values(
```
nick-gorman, line 212: Hey Ellie, 

Think I may have misunderstood why you just took the NSW column when I did the previous review. I didn't realise that the other columns were calculated column in the workbook.

Having said that, I wonder if this is one instance where we should be using the calculated column, so we can avoid assuming that NSW will always be the base values.

What are your thoughts, prehapes I'm being to pendantic?

nick-gorman:
Another minor comment. Should we make the generator name the combination of the technology and the rez? i.e. tech + rez. To make it unique for each row in the table. I tried to look if that was happening already but maybe I missed it.

EllieKallmier:
> Another minor comment. Should we make the generator name the combination of the technology and the rez? i.e. tech + rez. To make it unique for each row in the table. I tried to look if that was happening already but maybe I missed it.

We could do this for sure, my feeling would be to add a new column holding the tech+rez combination in case the raw technology for mappings to other features? 

Also noting the cases for which there is no rez (batteries, PHES, OCGT and CCGT) we could use subregion?

src/ispypsa/templater/static_new_generator_properties.py
EllieKallmier: Ahh yep all good! 

So at the moment I've actually separated out the opex "base" values from the opex LCFs and collected them separately into the summary table so that each could be edited separately. If we do want to just use the calculated values I would probably not collect the opex-specific LCFs to avoid confusion, which would simplify things a bit too. 

I'm happy to do this and just lose a tiny bit of that edit-ability here for the sake of simplicity and as you say avoid assumptions, let me know what you think!

nick-gorman:
> > Another minor comment. Should we make the generator name the combination of the technology and the rez? i.e. tech + rez. To make it unique for each row in the table. I tried to look if that was happening already but maybe I missed it.
> 
> We could do this for sure, my feeling would be to add a new column holding the tech+rez combination in case the raw technology for mappings to other features?
> 
> Also noting the cases for which there is no rez (batteries, PHES, OCGT and CCGT) we could use subregion?

Hi Ellie, yeah I agree keeping the raw tech for mapping is a good idea, and also using region instead of REZ where appropraite.

src/ispypsa/templater/static_new_generator_properties.py
nick-gorman: ok, that makes sense, I agree that it wouldn't make sense to include the factors doing it this way.

I guess overall, I'm probably pro using the calculated values in this instance.

EllieKallmier: Cool! That feels good to me, for the reasons you've outlined above. I'll get that happening shortly :)

EllieKallmier:
Latest changes (30/01):
- Added a new column that holds a combination of generator (full name) and either REZ (full name) or ISP subregion (abbreviation), depending on whether a REZ location existed for the generator.
 -> Used generator full names (as opposed to technology_type) to ensure uniqueness for each row
 -> Used REZ full names for consistency with values present in the new_entrants_summary table `rez_location` column.
- Edited the processing and merging of OPEX values to use the calculated values directly from each relevant table (fixed and variable opex). 
- Added fom and vom back to the `_NEW_GENERATOR_STATIC_PROPERTY_TABLE_MAP` and updated some of the control logic in `_merge_and_set_new_generators_static_properties()` function to handle this. 
 -> I did this to clearly show where the values come from in the case of workbook updates, so these tables are documented along with the other static property tables. 
 ->This also simplified the flow a bit by allowing for one function to handle both fixed and variable opex.

nick-gorman (approved):
Looks good Ellie!

I just made one sneaking change, hope thats ok.

PR #37, opened 2025-01-29
Rez build limits

nick-gorman:
Implements rez transmission limits
- simple transmission limits with extendable lines between rez and subregions are used for most rezs
- for some rezs, transmission limits are modelled using custom constraints that implement the "group constraints" and "transmission limit constraints" described in the IASR workbook


src/ispypsa/templater/renewable_energy_zones.py
```diff
@@ -47,3 +49,140 @@ def template_renewable_energy_zones(
...
+        ].isna() & (data[col] == 0.0)
+        data.loc[replacement_check, col] = np.nan
+    return data
+
+
+def combine_transmission_expansion_cost_to_one_column(data):
+    """The model can only utilise a single transmission expansion cost. If the tranche
+    1 column is nan then this function adopts the tranche 2 cost if it is not
+    nan. The process is repeated with tranche 3 if the cost is still nan.
+    """
```
EllieKallmier, line 169: Does there need to be any handling of a case where all three options are nan? It doesn't seem like this would cause any errors necessarily so I guess just wondering aloud here more than anything.

EllieKallmier (commented):
Things look good to me here! A few little questions more out of interest than anything else but otherwise looks solid, I think everything makes sense to me so far :)

src/ispypsa/templater/renewable_energy_zones.py
EllieKallmier: Actually I think this is kind of explained/handled in [PR#38 ](https://github.com/Open-ISP/ISPyPSA/pull/38), where the nan-ness of capital cost is used to set whether the line is extendable or not. So maybe already answered?

src/ispypsa/model/lines.py
```diff
@@ -4,6 +4,13 @@
 import pypsa
```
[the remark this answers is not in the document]
nick-gorman: Excellent point. This was a change that made sense when I was using 'dummy' lines for modelling group constraints, but it isn't needed now, so I'll combine them back into one function.

src/ispypsa/model/custom_constraints.py
```diff
@@ -0,0 +1,89 @@
...
+                network.model, row["variable_name"], row["component"], row["attribute"]
+            ),
+            axis=1,
+        )
+
+        # Some variables may not be present in the modeled so these a filtered out.
+        # variables that couldn't be found are logged in get_variables so this doesn't
+        # result in 'silent failure'.
+        retrieved_vars = ~variables.isna()
+        print(retrieved_vars)
```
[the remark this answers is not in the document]
nick-gorman: Ah, yes, that was just for debugging; I'll delete it. Thanks :)

src/ispypsa/templater/renewable_energy_zones.py
nick-gorman: Yep, that's correct, going to use a nan capital cost to allow users to turn expansion off for individual lines.

nick-gorman:
These changes were merged with #38 so will just close this PR.

PR #42, opened 2025-02-12
Template energy policy tables

emilyzzhao:
Added `template_energy_policy.py` to the templater module to create four new tables, `renewable_share_targets.csv` (trajectories for renewable energy %), `renewable_generation_targets.csv` (state and NEM-level trajectories for renewable energy MWh), `powering_australia_plan.csv` (AEMO trajectories for NEM-level renewable energy % by scenario), and `technology_capacity_targets.csv` (state and NEM-level trajectories for renewable, storage, and offshore wind tech MW). These tables are drawn from the AEMO trajectories which were interpolated from state and national energy policy targets. 

Adjusted `mappings.py` to map specific CSVs to the right functions, and adjusted dodo.py. 

These tables were aggregated mostly based on units and sort of inspired by how PyPSA-USA structured their policy inputs. I haven't written a test yet because wanted to check this was an appropriate way to structure the policy tables.


nick-gorman (commented):
Policy template tables look good! See a couple of minor comments.

Additionally, and I don't think this requires any changes, and is more just a thought I had while reading the code. I think it also would have been an ok approach to hard code the names of the target rows, rather than detecting them based on naming patterns. In some ways using naming patterns is more robust, but in another way its fragile because it relies on AEMO continuing to follow the pattern ;)

src/ispypsa/templater/energy_policy_targets.py
```diff
@@ -0,0 +1,228 @@
...
+        file_path = Path(parsed_workbook_path, f"{target['csv']}.csv")
+
+        df = pd.read_csv(file_path)
+        df = df.melt(id_vars=df.columns[0], var_name="FY", value_name="pct")
+        df = df[df[df.columns[0]].str.contains("share", case=False)]
+        df["region_id"] = target["region_id"]
+
+        # Convert percentage to decimal if needed
+        df["pct"] = df["pct"].astype(float)
+        df.loc[df["pct"] > 1, "pct"] = df["pct"] / 100
```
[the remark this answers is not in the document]
emilyzzhao: Makes sense! I'll change it to 0-100.

src/ispypsa/templater/energy_policy_targets.py
```diff
@@ -0,0 +1,228 @@
...
+    # Get mapping for this function
+    target_files = _TEMPLATE_RENEWABLE_ENERGY_TARGET_MAP[
+        "template_renewable_share_targets"
+    ]
+
+    for target in target_files:
+        file_path = Path(parsed_workbook_path, f"{target['csv']}.csv")
+
+        df = pd.read_csv(file_path)
+        df = df.melt(id_vars=df.columns[0], var_name="FY", value_name="pct")
```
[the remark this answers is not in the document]
emilyzzhao: Agreed we should figure out whether to have years as rows or columns because closure_years.csv and liquid_fuel_prices.csv have years as rows! For some of the policy tables, if years were columns, some regions might lack data for specific years (i.e. technology_capacity_targets.csv)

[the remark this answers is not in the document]
nick-gorman: That's a good point about the missing data for some regions. I guess that's another argument in favour of years as rows.

liquid_fuel_prices is a weird one where it's as columns, but because it's just one column, it gets converted to an index that is then transformed to rows when writing to CSV, or something like that.

emilyzzhao:
Hello, biggest new changes that have been made: 

- New `_template_energy_policy_targets` function which feeds into the `create_template` script. Groups together all the individual template policy table functions. 
- New policy_id table, `policy_generator_types.csv` which has the policy id as a column and the eligible generator types under that policy. I imagined this will come into play in the translation or model steps. One lingering question is I current have CST as "Concentrated Solar Thermal" but maybe should change to "Solar Thermal (15hrs storage)" to be consistent with New entrant data summary? 
- All policy tables now have policy_id column which links them with the policy_id table. 
- Tests adjusted to align with new api refactor

src/ispypsa/templater/manually_extracted_template_tables/6.0/policy_generator_types.csv
```diff
@@ -0,0 +1,47 @@
+﻿policy_id,generator
```
EllieKallmier, line 1: I have a question about Wind specifically here: when "Wind" is given as the generator here, is that only onshore wind (exclusive of any offshore)? Also think there might be a double up in row 3 and row 12 (cis_generator, Wind), but maybe this is related?

src/ispypsa/templater/manually_extracted_template_tables/6.0/policy_generator_types.csv
```diff
@@ -0,0 +1,47 @@
...
+vic_storage,Pumped Hydro (8hrs storage)
+vic_storage,Pumped Hydro (24hrs storage)
+vic_storage,Pumped Hydro (48hrs storage)
+vic_offshore_wind,Wind - offshore (fixed)
+vic_offshore_wind,Wind - offshore (floating)
+qret,Hydro
+qret,Large scale Solar PV
+qret,Wind
+qret,Biomass
+qret,Concentrated Solar Thermal
```
EllieKallmier, line 34: To your note/q about this naming, I reckon if we keep all the generator references named the same across all templater steps that might make our lives easier later on when we try to connect all the pieces in translation/model steps. So if not too annoying possibly we could use "Solar Thermal (15hrs storage)" here? Open to other thoughts ofc :)

EllieKallmier (approved):
Huge! Looks good to me, just a few small notes/questions left as comments here :)

src/ispypsa/templater/manually_extracted_template_tables/6.0/policy_generator_types.csv
emilyzzhao: yup good question, QRET only permits onshore wind (probably because they don't plan to have any offshore wind farms), but the rest of the policies allow offshore wind. So I guess the best format would be to have something like
cis_generator,Wind
cis_generator, Wind - offshore (fixed)
cis_generator, Wind - offshore (floating)

src/ispypsa/templater/manually_extracted_template_tables/6.0/policy_generator_types.csv
emilyzzhao: Yeah that sounds good!

src/ispypsa/templater/manually_extracted_template_tables/6.0/policy_generator_types.csv
EllieKallmier: Oh cool, yeah that looks good to me!

PR #51, opened 2025-09-11
Mkdocs setup & CLI implementation (etc etc)

nick-gorman:
##  Summary

  This PR adds:

- MkDocs documentation
- a command-line interface built on doit, 
- regional filtering capabilities for ISPyPSA input templates
- snapshot level weightings
- updates to test data sets (although trace data will need to change again)
- some other random reorganisation

## Command-Line Interface (CLI)

New ispypsa command-line interface built on doit:

  - Task structure: Six tasks with automatic dependency management: cache_required_iasr_workbook_tables, create_ispypsa_inputs,
  create_pypsa_friendly_inputs, create_and_run_capacity_expansion_model, create_operational_timeseries,
  create_and_run_operational_model
  - Optional optimization: dont_run_capacity_expansion and dont_run_operational flags to skip optimization steps

 ## Regional Filtering Feature

  Added capability to filter ISPyPSA templates by region or sub-region:

  - Implementation: New filter_template.py module for filtering by NEM regions or sub-regions
  - Functionality: Reduces model scope to specified geographic areas

  ## Translator & Snapshots Changes

  Updates to PyPSA translation and snapshot handling:

  - Snapshot weighting: Added _add_snapshot_weightings function
  - Custom constraints: Extracted custom constraint logic to separate module (custom_constraints.py, 30 lines)

  ## Testing Data Updates

  Updated test IASR data and trace files:

  - Update test IASR data table set. Inputs in tables have stripped back to minimum needed for testing.
  - Trace data: Restructured renewable and demand traces

### DIFF
--- a/src/ispypsa/config/validators.py
+++ b/src/ispypsa/config/validators.py
@@ -170,4 +170,3 @@ class ModelConfig(BaseModel):
                 "Cannot specify both filter_by_nem_regions and filter_by_isp_sub_regions"
             )
         return self
-


docs/config.md
```diff
@@ -0,0 +1,366 @@
...
+Weighted average cost of capital for annuitisation of generation and transmission costs,
+as a fraction, i.e. 0.07 is 7%.
+
+Examples:
+
+```wacc: 0.07```
+
+### discount_rate
+
+Discount rate applied to model objective function, as a fraction, i.e. 0.07 is 7%.
```
EllieKallmier:
```suggestion
Discount rate applied to model objective function, as a fraction, i.e. 0.05 is 5%.
```

maybe - just to line up with example + be different to wacc definition

docs/method.md
```diff
@@ -0,0 +1,249 @@
...
+### REZ exports
+
+REZ exports are implemented with a single static limit on power flow that applies in
+both directions of flow:
+
+- The value `rez_transmission_network_limit_summer_typical` from the ISPyPSA input table
+`renewable_energy_zones` is used to set power flow limit.
+- TODO: Implement time varying export limits making use of the peak demand and
+winter limits provided in the IASR workbook.
+- If a NA of blank value is provided then the `rez_to_sub_region_transmission_default_limit`
```
EllieKallmier:
```suggestion
- If a NA or blank value is provided then the `rez_to_sub_region_transmission_default_limit`
```

docs/method.md
```diff
@@ -0,0 +1,249 @@
...
+Weather reference years are used ensure weather correlations are consistent between demand
+and renewable energy availability, and to ensure the modelling considers diverse weather
+conditions:
+
+- For each model year between the model start year and model end year, a historical reference year
+is chosen from which to derive both demand and renewable energy availability.
+- In the model config the inputs [temporal.capacity_expansion.reference_year_cycle](config.md#temporalcapacity_expansionreference_year_cycle)
+and [temporal.operational.reference_year_cycle](config.md#temporaloperationalreference_year_cycle)
+are used to specify the ordering of reference years.
+- If the `reference_year_cycle` is shorter than model the horizon than cycle is repeated as many
```
EllieKallmier:
```suggestion
- If the `reference_year_cycle` is shorter than the model horizon then the cycle is repeated as many
```

src/ispypsa/templater/filter_template.py
```diff
@@ -0,0 +1,452 @@
...
+        invalid_regions = set(isp_sub_regions) - set(valid_sub_regions)
+        for region in invalid_regions:
+            logging.warning(f"ISP sub-region '{region}' not found")
+
+        # Filter to valid sub-regions
+        selected_sub_regions = [r for r in isp_sub_regions if r in valid_sub_regions]
+
+    if not selected_sub_regions:
+        logging.warning("No valid regions found for filtering")
+        return [], []
```
EllieKallmier: maybe similar question to above (just a q for me) - is it generally preferred/better practice to e.g. return values here that later (after this function has been called and run) trigger an error rather than raising the error here? 

I'm imagining it's probably in conjunction with logging helpful for tracing the error back if it's getting raised in the higher-level function?

tests/test_workbook_table_cache/connection_costs_other.csv
```diff
@@ -1,6 +1,2 @@
-Region,CCGT,CCGT with CCS,Small OCGT,Large OCGT,Reciprocating engines,Hydrogen Reciprocating engines,Biomass,1 hr Battery Storage,2 hr Battery Storage,4 hr Battery Storage,8 hr Battery Storage,Pumped Hydro3 (8 hrs storage),Pumped Hydro (24 hrs storage),Pumped Hydro (48 hrs storage),BOTN- Cethana
-QLD,110.49433333333297,110.49433333333297,110.49433333333332,110.49433333333332,110.49433333333332,110.49433333333332,110.49433333333332,103.36566666666667,103.36566666666667,103.36566666666667,103.36566666666667,0,0,0,0
+Region,CCGT,CCGT with CCS,Small OCGT2,Large OCGT,Reciprocating engines,Hydrogen Reciprocating engines,Biomass,1 hr Battery Storage,2 hr Battery Storage,4 hr Battery Storage,8 hr Battery Storage,Pumped Hydro3 (8 hrs storage),Pumped Hydro (24 hrs storage),Pumped Hydro (48 hrs storage),BOTN- Cethana
```
EllieKallmier: looks like footnote back here too? (Matching?)

docs/config.md
```diff
@@ -0,0 +1,366 @@
...
+
+Examples:
+
+```resolution_min: 30```
+
+#### temporal.capacity_expansion.reference_year_cycle
+
+The order in which different weather reference years are used in the model. If the
+number of reference years is less than the number of model years the reference years
+will be reused in the order provided.
```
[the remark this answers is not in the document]
nick-gorman: Yep, great idea!

src/ispypsa/templater/filter_template.py
```diff
@@ -0,0 +1,452 @@
...
+                        Cannot be specified together with nem_regions.
+
+    Returns:
+        Dictionary of filtered dataframes
+
+    Raises:
+        ValueError: If both or neither of nem_regions and isp_sub_regions are provided
+    """
+    # Validate inputs
+    if (nem_regions is None) == (isp_sub_regions is None):
```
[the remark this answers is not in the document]
nick-gorman: hmmmm it's a good question. It's a tricky one, because on one hand I'm very much in favour of just doing checks like this once to avoid clutter. And the config validation is the logical place to group all the checks like this together in one place. On the other hand create_ispypsa_inputs_template where the filtering is done is designed to be used without needing a config, so the checks at the config stage could be missed. If I had to right rules they would be:

- If the config will always catch it, only put the check in the config (for example the translator takes a config object so everything in there is protected by the config checks).
- If the config won't always offer protection there should be checks to cover the case where the config will miss it. But I'd still keep the config check because its nice to know up front that a config is completely vaild.

not saying I have followed these rules :)

src/ispypsa/templater/filter_template.py
nick-gorman: OK yep this is probably bad practice. It would be simpler (and therefore better) to just raise the here.

PR #56, opened 2025-11-19
Generator translation and handling

EllieKallmier:
This PR sets up translation of both ECAA and new entrant generators from `ISPyPSA` formatted tables (outputs of the `templater` step) into `PyPSA`-friendly formatted tables. Main features of this PR and the Generator translation include:

- Adding investment periods as build years and creating new entrant generators of each type/REZ/subregion combination for each possible build year
- Calculating and setting new entrant generator capital costs (annuitised). Includes consideration of build-year-specific build costs (locational), connection costs and associated system strength costs where applicable.
- Calculating dynamic marginal costs for each unique generator type/fuel/region/subregion/REZ combination and storing them as timeseries files to add to the model. Functionality also includes ability to manually over-write the string value in `marginal_cost` column with a numeric value to set a static marginal cost for specific generators.
- Fetching and adding new entrant generator traces to the network
- Implementation of build and resource limits as custom constraints, with dummy generators dynamically created to relax resource limit constraints where appropriate.

### DIFF
--- a/src/ispypsa/templater/static_new_generator_properties.py
+++ b/src/ispypsa/templater/static_new_generator_properties.py
@@ -5,12 +5,16 @@ import pandas as pd
 
 from .helpers import (
     _fuzzy_match_names,
+    _manual_remove_footnotes_from_generator_names,
     _one_to_one_priority_based_fuzzy_matching,
+    _rez_name_to_id_mapping,
     _snakecase_string,
+    _standardise_storage_capitalisation,
     _where_any_substring_appears,
 )
-from .lists import _NEW_GENERATOR_TYPES
+from .lists import _MINIMUM_REQUIRED_GENERATOR_COLUMNS, _NEW_GENERATOR_TYPES
 from .mappings import (
+    _NEW_ENTRANT_GENERATOR_NEW_COLUMN_MAPPING,
     _NEW_GENERATOR_STATIC_PROPERTY_TABLE_MAP,
     _VRE_RESOURCE_QUALITY_AND_TECH_CODES,
 )
@@ -76,7 +80,13 @@ def _template_new_generators_static_properties(
         .reset_index(drop=True)
     )
 
-    return new_generator_summaries
+    required_cols_only = [
+        col
+        for col in _MINIMUM_REQUIRED_GENERATOR_COLUMNS
+        if col in new_generator_summaries.columns
+    ]
+
+    return new_generator_summaries[required_cols_only]
 
 
 def _clean_generator_summary(df: pd.DataFrame) -> pd.DataFrame:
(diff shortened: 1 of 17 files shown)


tests/test_workbook_table_cache/connection_costs_other.csv
nick-gorman: Any idea what's happening here? I'm not sure why the footnote for OCGT is back.

src/ispypsa/translator/generators.py
```diff
@@ -1,60 +1,998 @@
...
         gen_attributes["sub_region_id"] = "bus"
+        bus_column = "sub_region_id"
     elif regional_granularity == "nem_regions":
         gen_attributes["region_id"] = "bus"
+        bus_column = "region_id"
+
+    if regional_granularity == "single_region":
+        ecaa_generators["bus"] = "NEM"
+        gen_attributes["bus"] = "bus"
+        bus_column = "bus"
```
nick-gorman: This is really trival, but maybe this is a bit clearer?

```suggestion

    # Decide which column to rename to be the bus column.
    if regional_granularity == "sub_regions":
        bus_column = "sub_region_id"
    elif regional_granularity == "nem_regions":
        bus_column = "region_id"
    elif regional_granularity == "single_region":
        # No existing column to use for bus, so create a new one.
        ecaa_generators["bus"] = "NEM"
        bus_column = "bus" # Name doesn't need to change.
    gen_attributes[bus_column] = "bus"
```

src/ispypsa/translator/generators.py
```diff
@@ -1,60 +1,998 @@
...
+            rez_mask, "rez_id"
+        ]
+
+    ecaa_generators["commissioning_date"] = ecaa_generators["commissioning_date"].apply(
+        _get_commissioning_or_build_year_as_int,
+        default_build_year=investment_periods[0],
+        year_type=year_type,
+    )
+    # Add marginal_cost col with a string mapping to the name of parquet file
+    ecaa_generators["marginal_cost"] = ecaa_generators["generator"].apply(
```
nick-gorman: Should we convert the other time series datafiles to also use this naming convention?

src/ispypsa/templater/static_new_generator_properties.py
```diff
@@ -380,19 +413,184 @@ def _zero_solar_wind_battery_partial_outage_derating_factor(
     return df
 
 
-def _add_technology_rez_subregion_column(
-    df: pd.DataFrame, new_col_name: str
+def _add_and_clean_rez_ids(
```
nick-gorman: Should REZs in the template use IDs instead of names? At the moment we are using ID in some tables and names in others, this will probably be pretty confusing for users. Other locations we are using names at the moment are: connection_cost_rez/_region_id, build_limit_region_id, new_entrant_wind_and_solar_connection_costs.csv

nick-gorman (commented):
Yo Ellie! very nice stuff! So cool to have all this generator functionality ready to go.

I definitely can't say I've fully had time to follow all the data manipulation magic you've had to do to make this work, but I think I'm at the point where I've captured 90%+ of the comments I'm going to make so better to submit the review rather than drag it out for little tidbits!

I've definitely used this as an opportunity to brain dump and some of the ideas are much broader than the scope of the PR so please take that into account when reading. Pardon the scattered nature and wording.

Have a read of my comments see what you think is worth addressing. Can always chat about it if that's helpful.

I think the first dot point below, about hiding tables and columns with no link to the model formulation, might be the one I'd be most keen to address before the release/workshop. Whether this PR or not probably isn't a huge deal, but could be very good to get your help with, as you know the generator stuff the best.

General comments 

- I 'think' there may be quite a few template tables and columns (particularly in ecca_generators and new_entrants_generators), which don't get incorporated in to the model formulation at the moment. If that is the case, I think we should temporarily stop them being output to the template. I can imagine it being quite confusing for the user if changing these values has no effect on the model. Maybe this PR isn't the place to make that change, idk.
- I think the order columns are displayed in ecca_generators and new_entrants_generators could be tweaked. Particularly bring id columns such as rez_id and generator over to the left.
- I think you may have explained this to me before, but just wanted to check in both the generator_name and technology type columns are needed in new_entrants_generators
- Definitely out of scope for the PR, but I wonder if we should drop the 2025-26_<unit> type notation from the template tables, in favor of 2026 and shift the units to the docs. Reading the PR I can see a bunch of places where this would simplify the codebase. It would also make fy and calendar year support much simpler.

PR #78, opened 2026-02-12
Fix CI/CD workflow to use correct Python versions and PyPI URL

nick-gorman:
- Pass matrix python-version to setup-uv, uv sync, and uv run so tests actually run against the specified Python versions
- Remove Python 3.9 from matrix (requires-python is >=3.10)
- Fix PyPI publish URL from isp-workbook-parser to ispypsa
- Bump actions/checkout to v4 and astral-sh/setup-uv to v5


dylanjmcconnell:
Hey Nick - given upgrade to `pypsa` to 1.0, the issue identified above would be addressed (namely that an updated version of `shapely` can be installed, i.e. one that has 3.14 wheels, and will works with the updated `pypsa`.)

I think given this (and that this PR is upstream of the 'pyspa-upgrade' PR) - think it makes most sense to merge this PR (then rebase `upgrade-pypsa` onto the updated main),  then add 3.14 to the `upgrade-pypsa` PR*

(*Haven't had a chance to properly look at that - seems okay, but maybe need to have  a better look / review too)

PR #93, opened 2026-04-14
Add initial set of generation and storage schemas

EllieKallmier:
Adds YAML validation schemas for generation and storage input tables to the initial set of ISPyPSA input table schemas.

Schemas added:
* `generators_existing_planned.yaml` - Existing and planned generator characteristics
* `generators_new_entrant.yaml` - New entrant generator technology options
* `storage_existing_planned.yaml` - Existing and planned storage unit characteristics
* `storage_new_entrant.yaml` - New entrant storage technology options (battery, PHES)
* `costs_connection.yaml` - Connection cost data
* `costs_fuel_prices.yaml` - Fuel price projections
* `costs_new_entrant_build.yaml` - Build cost projections for new entrants
* `emissions_reduction.yaml` - Emissions reduction targets/constraints

Still to be added: policy table schemas. 

Schemas follow structure as outlined in #85 and mostly follow the draft table structures given in the [alternative templater review](https://github.com/Open-ISP/project-planning/pull/4/changes), with tweaks to match the 2026 (v7.5) IASR workbook data.

### DIFF
--- a/src/ispypsa/validation/schemas/costs_connection.yaml
+++ b/src/ispypsa/validation/schemas/costs_connection.yaml
@@ -1,0 +1,62 @@
+table: costs_connection
+required: false
+unique:
+  - [geo_id, technology, year]
+description: >
+  Connection costs and system security costs for new entrant generators and storage.
+
+  Contains additional costs for grid connection and, where applicable, system
+  strength requirements for inverter-based resources.
+
+  Source tables:
+    - `connection_cost_forecast_wind_and_solar`
+    - `connection_cost_forecast_other`
+
+  If absent: no additional connection or system security costs apply to new entrant
+  build costs.
+columns:
+  geo_id:
+    type: string
+    required: true
+    allowed_values_from:
+      table: network_geography
+      column: geo_id
+    description: >
+      Network geography identifier where the connection cost applies.
+
+      If costs are given by region in source data, create separate rows for each
+      geo_id in the region with the same cost.
+  technology:
+    type: string
+    required: true
+    description: >
+      Standardised technology name mapping to new entrant technologies in
+      `generators_new_entrant` and/or `storage_new_entrant` tables.
+
+      If blank: treat as geo_id-level VRE connection cost.
+  year:
+    type: int
+    required: true
+    description: >
+      Financial year in which this cost applies.
+
+      Used to map to new entrant storage/generation by build year.
+  connection_cost:
+    type: float
+    required: true
+    units: $/MW
+    description: >
+      Additional cost for new build reflecting transmission infrastructure required
+      for grid connection (AUD). Must be >= $0.0.
+  system_strength_cost:
+    type: float
+    required: false
+    units: $/MW
+    description: >
+      Additional cost for inverter-based resources to meet efficient levels of
+      system strength (AUD). Must be >= $0.0 or blank (assumed $0.0).
+
+      Per IASR: "This cost is a blend of remediation with synchronous condenser
+      and grid forming batteries".
+
+      If absent: only `connection_cost` is added to new entrant build cost.
--- a/src/ispypsa/validation/schemas/costs_connection.yaml
+++ b/src/ispypsa/validation/schemas/costs_connection.yaml
@@ -11,52 +11,67 @@ description: >
   Source tables:
     - `connection_cost_forecast_wind_and_solar`
     - `connection_cost_forecast_other`
+    - `connection_costs_for_wind_and_solar`
+    - `connection_costs_other`
+    - `efficient_level_of_system_strength_cost`
 
-  If absent: no additional connection or system security costs apply to new entrant
-  build costs.
+  If absent:
+  No additional connection or system security costs apply to new entrant build costs.
 columns:
   geo_id:
     type: string
     required: true
     allowed_values_from:
-      table: network_geography
-      column: geo_id
+      - network_geography: geo_id
     description: >
       Network geography identifier where the connection cost applies.
 
-      If costs are given by region in source data, create separate rows for each
-      geo_id in the region with the same cost.
+      Source notes:
+      If costs are given by region in source data, separate rows for each geo_id in
+      the region are created by the `templater` with the same cost applied across each.
   technology:
     type: string
-    required: true
+    required: false
+    allowed_values_from:
+      - costs_new_entrant_build: technology
     description: >
-      Standardised technology name mapping to new entrant technologies in
-      `generators_new_entrant` and/or `storage_new_entrant` tables.
+      Standardised technology name mapping to new entrant technologies.
 
-      If blank: treat as geo_id-level VRE connection cost.
+      If absent (or empty):
+      The connection cost applies to all new entrant technologies in `geo_id` that
+      have no technology-specific connection cost for the applicable year.
   year:
     type: int
     required: true
-    description: >
-      Financial year in which this cost applies.
-
-      Used to map to new entrant storage/generation by build year.
+    description: Year in which this cost applies.
   connection_cost:
     type: float
-    required: true
+    required: false
     units: $/MW
+    gte: 0.0
+    nan_fill: 0.0
     description: >
-      Additional cost for new build reflecting transmission infrastructure required
-      for grid connection (AUD). Must be >= $0.0.
+      Additional cost in AUD for new build reflecting transmission infrastructure required
+      for grid connection.
+
+      Used in calculating the total capital cost of new entrant technologies.
+
+      If absent (or empty):
+      No additional connection cost applies for the corresponding technology, geo_id
+      and build year.
   system_strength_cost:
     type: float
     required: false
     units: $/MW
+    gte: 0.0
+    nan_fill: 0.0
     description: >
-      Additional cost for inverter-based resources to meet efficient levels of
-      system strength (AUD). Must be >= $0.0 or blank (assumed $0.0).
+      Additional cost in AUD for inverter-based resources to meet efficient levels of
+      system strength.
 
-      Per IASR: "This cost is a blend of remediation with synchronous condenser
-      and grid forming batteries".
+      Used in calculating the total capital cost for some new entrant technologies.
+      Per IASR, "This cost is a blend of remediation with synchronous condenser
+      and grid forming batteries" (IASR workbook v7.5, 'Power System Security' sheet).
 
-      If absent: only `connection_cost` is added to new entrant build cost.
+      If absent (or empty):
+      Only the `connection_cost` value is added to new entrant build costs.
--- a/src/ispypsa/validation/schemas/costs_fuel_prices.yaml
+++ b/src/ispypsa/validation/schemas/costs_fuel_prices.yaml
@@ -1,0 +1,47 @@
+table: costs_fuel_prices
+required: false
+unique:
+  - [fuel_type, fuel_price_node, year]
+description: >
+  Fuel prices by year for generation and storage units.
+
+  Contains fuel prices used to calculate dynamic marginal costs for thermal generators.
+
+  Source tables:
+    - `coal_fuel_price`
+    - `biomass_fuel_price`
+    - `gas_and_liquid_fuel_prices_consultant_scenario_mapping`
+    - `gas_prices_existing_generators`
+    - `gas_prices_new_entrants`
+    - `liquid_fuel_prices`
+    - `gpg_secondary_fuel_prices` # maybe
+    - `hydrogen_prices`
+    - `biomethane_prices`
+  The consultant scenario mapping table is needed to map between specified ISP
+  scenario names and the gas, liquid fuel, hydrogen and biomenthane price scenario
+  names. This is new in the 2026 ISP (IASR v7.5).
+
+  If absent: no dynamic marginal costs calculated. Requires later user input of
+  fixed or dynamic marginal cost.
+columns:
+  fuel_type:
+    type: string
+    required: true
+    description: Fuel type used by the generator or storage unit.
+  fuel_price_node:
+    type: string
+    required: true
+    description: >
+      Identifier mapping to specific assets or sets of new entrant assets.
+
+      For existing and planned generation/storage, maps to specific power stations.
+      For new entrants, maps to technology/location groups.
+  year:
+    type: int
+    required: true
+    description: Financial year for which the price applies.
+  price:
+    type: float
+    required: true
+    units: $/GJ
+    description: Fuel price in AUD per gigajoule.
--- a/src/ispypsa/validation/schemas/costs_fuel_prices.yaml
+++ b/src/ispypsa/validation/schemas/costs_fuel_prices.yaml
@@ -1,5 +1,12 @@
 table: costs_fuel_prices
-required: false
+# note: this table is required to calculate SRMC. Possible to make this optional in
+# future and/or design in more flexibility around dynamic vs. static SRMC calculation
+# - but I'm getting blocked on how to implement this at the moment with some
+# complexity around VRE vs non-VRE data requirements and how validation will
+# be applied (to make some decisions around how conditional requirements can
+# work) - basically needs more thought/discussion so for now I'm making all
+# columns/tables needed to calculate SRMC (dynamic) required.
+required: true
 unique:
   - [fuel_type, fuel_price_node, year]
 description: >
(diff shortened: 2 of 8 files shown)


src/ispypsa/validation/schemas/costs_connection.yaml
```diff
@@ -0,0 +1,62 @@
...
+    type: string
+    required: true
+    allowed_values_from:
+      table: network_geography
+      column: geo_id
+    description: >
+      Network geography identifier where the connection cost applies.
+
+      If costs are given by region in source data, create separate rows for each
+      geo_id in the region with the same cost.
```
nick-gorman: Great idea to have this note. Do you think we should consistently use a heading like "Source notes", "Data preparation notes" to preface notes like this. Anyway not a big one, just a thought.

src/ispypsa/validation/schemas/costs_connection.yaml
```diff
@@ -0,0 +1,62 @@
...
+  technology:
+    type: string
+    required: true
+    description: >
+      Standardised technology name mapping to new entrant technologies in
+      `generators_new_entrant` and/or `storage_new_entrant` tables.
+
+      If blank: treat as geo_id-level VRE connection cost.
```
nick-gorman: Should this be more explicit. Maybe, "If blank: the cost is applied to all new entrant technologies in geo_id which have not been provided a technology specific connection cost for the applicable year."

src/ispypsa/validation/schemas/costs_connection.yaml
```diff
@@ -0,0 +1,62 @@
...
+  year:
+    type: int
+    required: true
+    description: >
+      Financial year in which this cost applies.
```
nick-gorman: Maybe just "Year" to leave open the cost being applied on calendar year basis.

src/ispypsa/validation/schemas/costs_fuel_prices.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - `liquid_fuel_prices`
+    - `gpg_secondary_fuel_prices` # maybe
+    - `hydrogen_prices`
+    - `biomethane_prices`
+  The consultant scenario mapping table is needed to map between specified ISP
+  scenario names and the gas, liquid fuel, hydrogen and biomenthane price scenario
+  names. This is new in the 2026 ISP (IASR v7.5).
+
+  If absent: no dynamic marginal costs calculated. Requires later user input of
+  fixed or dynamic marginal cost.
```
nick-gorman: Should we be more specific about where these inputs would need to be provided if they aren't given here?

src/ispypsa/validation/schemas/costs_new_entrant_build.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+  Source tables:
+    - `build_costs`
+
+  Conditionally required: if new entrant generation or storage asset tables are
+  present and non-empty, build costs must be provided.
+columns:
+  technology:
+    type: string
+    required: true
```
nick-gorman, line 19: Should we include the following:

```yaml
allowed_values_from:
  - generators_new_entrant: technology
  - storage_new_entrant: technology
```

Note: this is a different format for allowed_values_from then I used with the network tables, but it allows for multiple tables, so maybe we can just update the network tables to use this more generic format. Let me know what you think.

src/ispypsa/validation/schemas/costs_connection.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+      Used to map to new entrant storage/generation by build year.
+  connection_cost:
+    type: float
+    required: true
+    units: $/MW
+    description: >
+      Additional cost for new build reflecting transmission infrastructure required
+      for grid connection (AUD). Must be >= $0.0.
```
nick-gorman: This should be inforced with validation?

src/ispypsa/validation/schemas/emissions_reduction.yaml
```diff
@@ -0,0 +1,50 @@
...
+  If absent: no fuel blending assumed, so no adjustments to fuel price or emissions
+  intensity for gas generators.
+columns:
+  fuel_type:
+    type: string
+    required: true
+    description: >
+      Fuel type blended into the gas network.
+
+      Should match fuel_type values in `generators_existing_planned` or
```
nick-gorman: Should these be validated?

src/ispypsa/validation/schemas/costs_fuel_prices.yaml
```diff
@@ -0,0 +1,47 @@
...
+columns:
+  fuel_type:
+    type: string
+    required: true
+    description: Fuel type used by the generator or storage unit.
```
nick-gorman: Should we note here the model behaviour if the fuel_type (or maybe even fuel_type price_node combo) for a generator is missing? Is this allowed? Maybe its the same behaviour as if the whole table is missing but just applied to a singe generator. Should this be stated at the table description level, noting that if absent behaviour applies to the whole table and on row basis. And I guess this commet applies to other tables as well.

src/ispypsa/validation/schemas/generators_existing_planned.yaml
```diff
@@ -0,0 +1,146 @@
...
+  fuel_price_node:
+    type: string
+    required: false
+    description: >
+      The power station name or other identifier used to map fuel prices to the generator.
+
+      Note: for all power stations except for Kogan Gas, this should exactly match
+      the `power_station` column.
+
+      If absent: assume no dynamic marginal price is calculated for this model run.
```
nick-gorman: Does this mean zero cost or a fixed cost?

src/ispypsa/validation/schemas/storage_existing_planned.yaml
```diff
@@ -0,0 +1,141 @@
...
+      Discharging efficiency (%). Must be between 0.0 and 100.0.
+
+      For PHES with only round-trip efficiency, calculated as round_trip**(1/2)
+      assuming symmetric charge/discharge.
+  soc_max:
+    type: float
+    required: false
+    units: '%'
+    description: >
+      Maximum allowed state of charge (%). Must be between 0.0 and 100.0,
```
nick-gorman: validation?

src/ispypsa/validation/schemas/storage_existing_planned.yaml
```diff
@@ -0,0 +1,141 @@
...
+      Annual capacity degradation rate (%). Must be between 0.0 and 100.0.
+
+      Not currently applied in the model.
+
+      If absent: no degradation modelled.
+  commissioning_date:
+    type: date
+    required: false
+    description: >
+      Date when the storage unit begins operation. Format: %d/%m/%Y
```
nick-gorman: Validation?

nick-gorman (approved):
This looks great Ellie, just a few comments / questions.

- I made a bunch of comments on just one table, but they might apply in other cases. But this should be obvious.
- In the inline comments I suggested validating the fuel types in the fuel cost table off the generator tables. But could also make sense to do this in reverse, if the fuel cost table exists require that the generator table values exist in the fuel cost table. But then this raises the question of what happens when the fuel cost table isn't provided. Another option would be to have a canonical values yaml (or csv) and we validate off it for things like fuel types and technology types etc. This is a broader issue than fuel types but just using it as an exmaple.
-  Do you think we should have a standard way of separating descriptions notes as pertaining to source / IASR vs ISPyPSA behaviour.

src/ispypsa/validation/schemas/costs_connection.yaml
EllieKallmier: Yep I hear this - maybe also though I'd add a note (but maybe in the `description`?) that says something along the lines of "Year type is either financial or calendar year based on config" (phrased better). 

Side thought: I'm not sure how we want to handle converting data that's given as FY into calendar (or vice versa) when it's just a single value for each year? I don't think a big deal but good to take a standard approach

src/ispypsa/validation/schemas/costs_new_entrant_build.yaml
EllieKallmier: mmm I like that suggested format! And yes, I'll add that here. I think there are a few things to think about in terms of the order that tables pass through validation for cases like this, and as you mentioned for cases where tables are optional/not present etc...

src/ispypsa/validation/schemas/generators_existing_planned.yaml
EllieKallmier: I was thinking it would mean a fixed price (user to supply), or that users would need to provide dynamic prices (if that's desired). I can clarify this wording to be more specifically about the calculation of dynamic prices and the consequence/required action.

I'm also not totally convinced this is the best way to handle this scenario, I just think that this is a case with an obvious set of options that might be desirable to users AND can offer a simplification. I would be very happy to chat more about this case and whether there's a nicer way to implement!

src/ispypsa/validation/schemas/costs_connection.yaml
nick-gorman: On the side thought, agree, but I would kind of leave this up to the user. I.e. the templater doesn't have a calendar year option, but if the user wants to fill out the ISPyPSA tables with their own inpus they are free to specify them as calendar years.

src/ispypsa/validation/schemas/generators_existing_planned.yaml
nick-gorman: Maybe if fuel_price_node isn't supplied then the user needs to add a column called fuel_price, which would just specify one fixed price. Or you could allow numeric values in fuel_price_node which override the dynamic values, and just not allow NaN. The user can set to zero if thats what they want.

src/ispypsa/validation/schemas/storage_existing_planned.yaml
EllieKallmier: yes for sure - at the time I was thinking it might be easier to define column-level validation rules (in the schema) once we had a more solid plan for how the validation will be handled (i.e. if there's an existing package or smth we want to use). But yeah it's probably just as easy to define clearly now and update that as we go. I might in that case use the custom_validation attribute at the column level and start defining a suite of standard validations. Lmk if you have other ideas or preferences!

EllieKallmier: also - could use `allowed_values` but atm that's defined as a list of permitted values, which doesn't translate well here; we could instead update the definition of that attribute.

src/ispypsa/validation/schemas/costs_fuel_prices.yaml
EllieKallmier: I have also been wondering about this, and I think it's still for me a bit of a question around whether what I've sort of proposed for the missing table case is the best way forward. But yes as it stands I think it makes sense to add some cross-validation here to the summary tables and as you say add a table-level note re: if absent behaviour.

src/ispypsa/validation/schemas/storage_existing_planned.yaml
nick-gorman: Ok, yep, I agree it could make sense to wait till we know more about how we will be doing the validation.

EllieKallmier:
# Collection of my (edited) thoughts and updates:

### New schema additions for validation metadata

Super open to not doing this/changing stuff around - particularly as validation “enforcement” gets set up and might require certain structures etc.. But to summarise - shift away from embedding validation constraints as prose in `description` fields (e.g. "All values should be >= $0.0") and towards explicit fields at the column level. The new fields I’m testing out are:

| Field | Meaning |
| --- | --- |
| `gte` | Greater than or equal to (inclusive lower bound) |
| `gt` | Strictly greater than (exclusive lower bound) |
| `lte` | Less than or equal to (inclusive upper bound) |
| `lt` | Strictly less than (exclusive upper bound) |
| `format` | Format string for date or string pattern validation (e.g. `"%d/%m/%Y"`) |
| `allowed_values` | Explicit list of permitted values |
| `allowed_values_from` | Reference to a column in another table whose values define the permitted set |
| `nan_fill` | Value applied to null/NaN cells at the validation enforcement step - IF `required : false` for this column. |

The intent is that these fields can be read directly by validation enforcement code without parsing prose. Table-level `custom_validation` is reserved for more complex cross-table/cross-column conditions that can't be expressed this way, and `allowed_values` stays as a static list (discrete). The idea being that these fields only exist as needed per column.

---

### Clarified semantics of `required` vs `nan_fill`

The column-level `required` field defines: **whether a column can contain null values at validation time (without raising an error).**

For columns that are `required: false` but have a sensible default value, `nan_fill` specifies the value that will be applied to any null/NaN cells during the validation enforcement step. For example:

`lcf_build:
  type: float
  required: false
  units: '%'
  gte: 0.0
  nan_fill: 100.0`

This says: the column is optional; if cells are null, fill them with 100.0 (i.e., no locational scaling); validate that all resulting values are >= 0.

---

### Standardised description sub-headers

Description fields now use consistent sub-headers to distinguish different types of content:

- **`Source tables:`** — the IASR workbook tables drawn on by the `templater` to produce this output table
- **`Source notes:`** — templater transformation notes (what has already happened to get to this point; not enforced by the validator)
- **`If absent:`** — behavioural consequence when the entire table or column is not present
- **`If absent (or empty):`** — used where both the column being absent and individual null cells have the same consequence

I’m not locked in on having the source notes stuff here, maybe it’s more useful for me at this stage during the templater refactor but will become redundant/doubled as the templater documentation gets filled out? But the idea being that `Source notes:` defines stuff that has happened before reaching validation, in part documenting the templater behaviour but also in particular where new rows have been added that don’t exist in the IASR workbook. Happy to lose this subheading if it feels like a double up/not that useful!

---

### Simplified dynamic marginal cost data requirements

Basically - I was getting stuck on how to imply this kind of flexibly required if/then structure in the schema without having a more concrete picture of the validation and where it sits in the flow, and if/how we required users to manually fill some values at different points etc - so I made an executive decision to simplify and remove this optionality/flexibility for the moment so I can just keep moving!

Previously, the generator tables used an `optional_sets` field to indicate that `fuel_price_node` (now `fuel_price_mapping`), `vom`, and `heat_rate` were all required together or all absent. This approach was removed. These columns — along with `fom` — are now simply `required: true` in both new entrant generator and storage tables. 

The same change applies to `costs_fuel_prices`, which has been made `required: true` at the table level (was previously optional). This is noted in the schema with a comment explaining the intent to revisit optionality of SRMC-related data once the validation enforcement design is clearer.

I do think this is important to revisit just with a more serious think about the user interaction piece (where and how to enforce)!

---

### Column identifier rename: `fuel_price_node` → `fuel_price_mapping`

I renamed the column `fuel_price_node` in the generator tables and `costs_fuel_prices` back to `fuel_price_mapping`. Because - in the IASR v7.5 workbook structure: fuel prices are not specifically linked to a geographic node - for many generators they are identified by a mapping ID that doesn't correspond to a specific location. There’s no region/subregion/rez/location column in fuel price tables anymore basically!

---

### Cross-table references: new entrant technology validation

The `technology` column in both `generators_new_entrant` and `storage_new_entrant` now has:

`allowed_values_from:
  - costs_new_entrant_build: technology`

This enforces that any technology defined in the new entrant asset tables also has a build cost entry. 

**Open question for discussion:** This validates that new entrant assets have build costs, but it doesn't validate that all entries in `costs_new_entrant_build` correspond to technologies actually defined in the asset tables (which in my mind would be about catching stale or misspelled entries).

src/ispypsa/validation/schemas/costs_fuel_prices.yaml
```diff
@@ -0,0 +1,68 @@
...
+# note: this table is required to calculate SRMC. Possible to make this optional in
+# future and/or design in more flexibility around dynamic vs. static SRMC calculation
+# - but I'm getting blocked on how to implement this at the moment with some
+# complexity around VRE vs non-VRE data requirements and how validation will
+# be applied (to make some decisions around how conditional requirements can
+# work) - basically needs more thought/discussion so for now I'm making all
+# columns/tables needed to calculate SRMC (dynamic) required.
+required: true
+unique:
+  - [fuel_type, fuel_price_node, year]
```
nick-gorman:
```suggestion
  - [fuel_type, fuel_price_mapping, year]
```

src/ispypsa/validation/schemas/costs_fuel_prices.yaml
```diff
@@ -0,0 +1,68 @@
...
+  scenario names and the gas, liquid fuel, hydrogen and biomenthane price scenario
+  names. This is new in the 2026 ISP (IASR v7.5).
+
+  Rows are also added by the `templater` for VRE fuels (Wind, Solar) with a price
+  of $0.0 for all years.
+columns:
+  fuel_type:
+    type: string
+    required: true
+    allowed_values: ["Black Coal", "Brown Coal", "Liquid Fuel", "Gas", "Water", "Solar", "Wind", "Biomass", "Biomethane"]
```
nick-gorman: I'd did something similar in network_geography.yaml, but I wonder if hard coding like this is actually bad as it prevents people defining new fuel costs. Thinking again of a toy example type case where I might just want to define whatever fuels I like Coal, Hydrogen, etc 

Another way of doing this might be force the generation tables to only have fuel types that exist in costs_fuel_prices.

nick-gorman (approved):
Hey Ellie,

This looks really good, big fan.

I think there are some broader questions about how we implement referential integrity, which we haven't quiet answered, similar to your question on costs_new_entrant_build. But I think that is something we can keep thinking through and doesn't need to block this PR by any means. I might open a discussion on referential integrity.

src/ispypsa/validation/schemas/costs_fuel_prices.yaml
EllieKallmier: yeahh I thought about this too, and after the chat this morning about the role of validator totally agree hard-coding isn't ideal here. And yep I like that option - will implement!

PR #94, opened 2026-04-21
Add network_transmission_path_limits templater

nick-gorman:
**Summary**
- Adds new table: network_transmission_path_limits
- Missing-data collapse. Paths with all-NaN capacity (REZs absent from initial_transmission_limits, or flow paths with blank cells) collapse to a single path_id-only row; the translator applies a default. Now also logged at WARNING so silently-defaulted paths are visible.
- Adds regional_granularity support to the new-format templater. nem_regions filters intra-region flow paths and re-keys cross-region ones; single_region drops flow paths and retargets REZ paths to a single "NEM" geo. Schema loosened: subregion_id is optional and "NEM" is a valid region_id.
- Adds a Logging section to CLAUDE.md codifying when each level is appropriate.
- Heads-up: this PR also refactors the paths templater that just landed in the previous PR. transmission_paths.py is folded into transmission.py because paths and limits share IASR sources and most of the extraction helpers are now cross-cutting. Apologies for putting you through a second pass here — in hindsight the paths + limits work should have been a single PR.

**Example output:**
```
  path_id   direction  timeslice         capacity
  CQ-NQ     forward    peak_demand       1200
  CQ-NQ     reverse    winter_reference  1910
  Q1-NQ     forward    peak_demand       750
  Q1-NQ     reverse    peak_demand       750
  Q3-NQ     (NaN)      (NaN)             (NaN)    ← collapsed: no explicit limits
```
**Where the changes live:**
```
  src/ispypsa/
  ├── templater/
  │   ├── transmission.py            ← new; orchestrator + flow/REZ extraction + collapse + granularity aggregation (replaces transmission_paths.py)
  │   ├── geography.py               ← granularity-aware (sub_regions / nem_regions / single_region)
  │   ├── mappings.py                ← _SINGLE_REGION_ID = "NEM"
  │   ├── transmission_paths.py      ← removed
  │   └── create_template.py         ← wires limits + granularity through to both templaters
  ├── iasr_table_caching/
  │   └── local_cache.py             ← caches initial_transmission_limits
  └── validation/schemas/
      ├── network_geography.yaml             ← subregion_id optional; region_id allows "NEM"
      └── network_transmission_paths.yaml    ← geo codes vary with granularity

  tests/
  ├── test_templater/
  │   ├── test_transmission.py                      ← full-frame content + collapse + log coverage
  │   ├── test_transmission_nem_regions.py          ← new
  │   ├── test_transmission_single_region.py        ← new
  │   ├── test_geography.py                         ← extended for the two new granularities
  │   ├── test_transmission_paths.py                ← removed
  │   └── test_create_ispypsa_inputs_template.py    ← integration test, wiring-only (presence + columns + row count)
  └── test_iasr_table_caching/
      └── test_local_cache.py                       ← updated cache table list

  CLAUDE.md                          ← new ## Logging section
  pyproject.toml                     ← isp-workbook-parser ≥ 2.8.0 (IASR v7.5)
```


src/ispypsa/templater/geography.py
```diff
@@ -1,34 +1,109 @@
...
 
 
 def _template_network_geography(
     sub_regional_reference_nodes: pd.DataFrame,
     renewable_energy_zones: pd.DataFrame,
+    regional_granularity: str,
 ) -> pd.DataFrame:
     """Creates the network_geography table from IASR workbook tables.
 
+    The shape of the output depends on ``regional_granularity``:
```
EllieKallmier, line 13: Might be nice for completeness to state here that all outputs have the `geo_id`, `geo_type` and `region_id` columns - contextualises the inclusion/exclusion of `subregion_id` column a bit more maybe?

tests/test_templater/test_transmission_nem_regions.py
```diff
@@ -0,0 +1,231 @@
...
+        N1-NSW,   N1,        NSW,     AC
+    """)
+    pd.testing.assert_frame_equal(
+        paths.reset_index(drop=True), expected_paths.reset_index(drop=True)
+    )
+
+    # Collapsed row: path_id only, no direction/timeslice/capacity.
+    assert len(limits) == 1
+    assert limits.iloc[0]["path_id"] == "N1-NSW"
+    assert pd.isna(limits.iloc[0]["direction"])
```
EllieKallmier: Maybe this is more a validation question than a testing question, but how much is it worth specifically checking that all three columns (direction/timeslice/capacity) that are expected to be NaN are returned here (with NaN values)?

tests/test_templater/test_transmission.py
```diff
@@ -0,0 +1,552 @@
...
+            flow_paths,
+            initial_limits,
+            renewable_energy_zones,
+            _empty_sub_regional_geography(),
+            "sub_regions",
+        )
+
+    assert "REZs absent from initial_transmission_limits" in caplog.text
+    assert "N3" in caplog.text
+    assert "Q1" not in caplog.text.split("REZs absent")[1]
```
EllieKallmier: possibly silly question - why check the split here (but not the line above)?

EllieKallmier: (I guess in case Q1 appears elsewhere in the log before the warning message) - but would that be likely/possible to happen during testing?

EllieKallmier (approved):
Looks solid! Some extremely tiny comments and ofc a few questions cause I can't help myself. I agree it seems to make quite a bit of sense to have paths and limits treated together, and the regional granularity handling here feels simple and easy too (nice). 

There are a few 'style' things I guess I've noticed while reviewing this pr and I wanted to check my understanding/interpretation of those for my own benefit going forward:
- Functions here use a kinda new/different docstring style with the "I/O Example" section (I like this) - is this a style you'd like to use from this point on? (Loosely - obv with usual flexibility to do what makes sense based on the function)
- With testing - each submodule test file has the all-caps column lists defined up top and I'm reading this as a kinda move towards making tests more obvious (instead of importing from a fixture/lists/mapping file in tests for example). I'm on board with this and will adopt going forward where it makes sense too!

tests/test_templater/test_transmission_nem_regions.py
nick-gorman: Yeah, good point, this should (and some othe tests) should just use the standard assert against a full dataframe definition.

tests/test_templater/test_transmission.py
nick-gorman: No good reason to be honest. Given the testing setup its not really needed. Might just change to one assert and make it explicit:

``` 
assert(
    "REZs absent from initial_transmission_limits "
    "(default will be applied downstream): ['N3']"
) in caplog.text
```

nick-gorman:
> Looks solid! Some extremely tiny comments and ofc a few questions cause I can't help myself. I agree it seems to make quite a bit of sense to have paths and limits treated together, and the regional granularity handling here feels simple and easy too (nice).
> 
> There are a few 'style' things I guess I've noticed while reviewing this pr and I wanted to check my understanding/interpretation of those for my own benefit going forward:
> 
> * Functions here use a kinda new/different docstring style with the "I/O Example" section (I like this) - is this a style you'd like to use from this point on? (Loosely - obv with usual flexibility to do what makes sense based on the function)
> * With testing - each submodule test file has the all-caps column lists defined up top and I'm reading this as a kinda move towards making tests more obvious (instead of importing from a fixture/lists/mapping file in tests for example). I'm on board with this and will adopt going forward where it makes sense too!

The new docstring style is just me trying something out to help make reading the code that claude rights a easier to understand. If it works for you too that's great! I'd say lets keep experimenting with it and see where we land??

The all-caps column lists wasn't as intentional, kind of just something that Claude did. I don't think its a big deal either way. Reflecting on it and thinking more broadly, as I write less code, and generating code takes less time, I'm defintely shifting to prioritising readability over DRY, as the main bottle neck is do I understand the code, rather than how long does it take to write (although DRY has maintainability benifits over just time to write).

nick-gorman:
And thinking about the fixtures vs constants in the local file further, because the file contains lost of instances where these columns names are hard coded, moving these lists to a fixtures file doesn't actually create a single change point for maintances. The lists at the top of the file just make a few places in the tests where empty dataframes are defined a bit more concise.

PR #102, opened 2026-05-07
Add new-format network_expansion templater

nick-gorman:
## Summary

- Adds a new-format `network_expansion` templater that turns IASR flow-path and REZ
  augmentation tables into two ISPyPSA inputs: `network_expansion_options` (selected
  least-cost option per expandable element) and `network_transmission_path_expansion_costs`
  (long-form $/MW cost trajectory).
- Wires the templater into `create_ispypsa_inputs_template` behind the existing
  `use_new_table_format` feature flag, with a granularity-aware filter that drops or
  re-keys augmentation entries when paths are aggregated to NEM regions / single region.
- Extends `_template_network_transmission` to inject zero-capacity parallel paths for
  augmentation corridors (e.g. `CNSW-SNW`) that exist alongside suffixed siblings
  (`CNSW-SNW_NTH`/`_STH`) — without this, the orchestrator misclassifies them as
  constraint relaxations.
- Partitions the IASR table cache by workbook version on disk (`6.0/`, `7.5/`) and
  drives required-table discovery from a checked-in `known_tables.yaml` manifest so the
  augmentation prefixes can be enumerated per version.
- Schema updates: `network_expansion_options` now keys on `(expansion_id, expansion_type)`
  with allowed values `forward`/`reverse`/`constraint_relaxation`; cost-per-MW divisor
  documented as `max(forward, reverse)`.
- Supporting changes: env-var override for feature flags so subprocess CLI tests can flip
  them, `_financial_year_string_to_end_year_int` helper, deduplicated fuzzy-match log
  lines, and CLAUDE.md additions covering I/O-example docstrings, integration-test
  scope, and the "no hidden preconditions" rule.

## Example output

`network_expansion_options` — physical paths emit forward+reverse rows; constraint
groups (ids not in `network_transmission_paths`) emit one `constraint_relaxation` row:

```
expansion_id  expansion_type          allowed_expansion  expansion_option
CQ-NQ         forward                 1000               Option 1   # asymmetric path: forward MW from selected option
CQ-NQ         reverse                 1200               Option 1   # ...and reverse MW from same option (least $/MW)
DN1-CNSW      forward                 500                Option 2a  # source had reverse_mw = NaN
DN1-CNSW      reverse                 0                  Option 2a  # NaN -> 0 (option provides no expansion this direction)
N1-NNSW       forward                 1660               Option 1   # REZ path: symmetric (forward == reverse)
N1-NNSW       reverse                 1660               Option 1
SWQLD1        constraint_relaxation   330                Option 1   # constraint group: single row, not in network_transmission_paths
```

`network_transmission_path_expansion_costs` — long-form, $/MW of the larger directional
capacity:

```
expansion_id  year  cost
CQ-NQ         2025  416666.67    # 500M / max(1000, 1200) = 500M / 1200 (asymmetric -> divisor is the larger side)
CQ-NQ         2026  425000.00    # next-year cost / same divisor (escalation visible across years)
N1-NNSW       2025  3539566.27   # 5.875B / 1660 (symmetric -> unambiguous divisor)
N1-NNSW       2026  3593401.81
SWQLD1        2025  1515.15      # 500k / 330 (constraint group: divisor is its own capacity)
```

## Where the changes live

```
src/ispypsa/
├── feature_flags.py                            # env-var overrides for subprocess tests
├── cli/dodo.py                                 # version-aware cache target list
├── iasr_table_caching/
│   ├── known_tables.yaml                       # NEW — per-version table manifest
│   └── local_cache.py                          # version-partitioned cache, prefix-driven aug discovery
├── templater/
│   ├── create_template.py                      # new-format branch wires in network_expansion
│   ├── network_expansion.py                    # NEW — orchestrator + helpers (~970 lines)
│   ├── transmission.py                         # _append_new_parallel_paths + flow_path_options arg
│   └── helpers.py                              # _financial_year_string_to_end_year_int, dedup'd fuzzy log
└── validation/schemas/
    ├── network_expansion_options.yaml          # expansion_type column, composite uniqueness
    └── network_transmission_path_expansion_costs.yaml  # cost-per-MW divisor doc

tests/
├── test_workbook_table_cache/
│   ├── 6.0/                                    # existing fixtures, moved
│   └── 7.5/                                    # NEW — fixtures for new-format pathway
├── test_templater/
│   ├── test_network_expansion.py               # NEW — ~870 lines, per-helper coverage
│   ├── test_transmission.py                    # parallel-path wiring
│   └── test_create_ispypsa_inputs_template.py  # integration wiring
├── test_cli/
│   ├── cli_test_helpers_new_table_formats.py   # NEW
│   └── test_create_ispypsa_inputs_new_table_formats.py  # NEW — end-to-end CLI run
└── test_iasr_table_caching/test_local_cache.py # version-partitioned cache assertions

scripts/build_75_test_cache.py                  # NEW — one-off to regenerate 7.5 fixtures
CLAUDE.md                                       # I/O example, integration test, hidden-precondition rules
```


src/ispypsa/templater/transmission.py
```diff
@@ -809,3 +840,132 @@ def _remap_limit_path_ids(
...
+            CNSW-SNW        forward    summer_typical    0
+            CNSW-SNW        forward    winter_reference  0
+            CNSW-SNW        reverse    peak_demand       0
+            CNSW-SNW        reverse    summer_typical    0
+            CNSW-SNW        reverse    winter_reference  0
+    """
+    new_paths, new_limits = _new_parallel_path_rows(
+        flow_path_options, set(paths["path_id"])
+    )
+    if new_paths.empty:
```
EllieKallmier: Is this check needed? Being picky to check my understanding of the data validation/handling empties practices - this one to me looks still a bit defensive - very happy to be wrong!

src/ispypsa/templater/network_expansion.py
```diff
@@ -0,0 +1,969 @@
...
+        flow_path_options["CQ-NQ"]:
+            Flow path  Option name     Fwd_MW  Rev_MW
+            CQ-NQ      CQ-NQ Option 1  1000    1200
+            CQ-NQ      CQ-NQ Option 2  text    text  # non-numeric, dropped
+
+        returns:
+            expansion_id  option_name     forward_mw  reverse_mw
+            CQ-NQ         CQ-NQ Option 1  1000.0      1200.0
+    """
+    if not flow_path_options:
```
EllieKallmier, line 334: (comparing to my comment in transmission.py) - this one makes sense to me re: making sure the right columns/structure is returned (where it otherwise wouldn't be without the check)

src/ispypsa/templater/network_expansion.py
```diff
@@ -0,0 +1,969 @@
...
+
+        returns:
+            N1-NNSW    # REZ mapped to its path_id
+            SWQLD1     # constraint group passes through
+            None       # NaN → None for caller to drop
+    """
+
+    def convert(r):
+        if pd.isna(r):
+            return None
```
EllieKallmier: what's the benefit of converting `NaN` to `None` here?

src/ispypsa/templater/network_expansion.py
```diff
@@ -0,0 +1,969 @@
...
+    long = df[[id_col, option_col] + year_cols].melt(
+        id_vars=[id_col, option_col], var_name="year", value_name="cost"
+    )
+    long = long.rename(columns={id_col: "expansion_id", option_col: "option_name"})
+    long["year"] = long["year"].map(_financial_year_string_to_end_year_int)
+    long["cost"] = pd.to_numeric(long["cost"], errors="coerce")
+    return long.dropna(subset=["cost"]).reset_index(drop=True)
+
+
+def _looks_like_financial_year(col: str) -> bool:
```
EllieKallmier: I feel like this func could also go into helpers.py for reuse

src/ispypsa/validation/schemas/network_expansion_options.yaml
```diff
@@ -1,11 +1,13 @@
...
-  - [expansion_id]
+  - [expansion_id, expansion_type]
 description: >
   Defines the selected expansion option and allowed capacity increase for
   expandable network elements. Covers both physical transmission paths and
-  group constraints that can be relaxed through expansion.
+  group constraints that can be relaxed through expansion. Physical paths
+  emit two rows (forward and reverse); constraint groups emit one
+  constraint_relaxation row.
 columns:
```
EllieKallmier, line 11: small deal - I've been switching up the `allowed_values_from` structure to use `- table_name: column_name` instead of splitting across `table: name \ column: name`, I think based on a suggestion from a previous review. I'm not particularly attached to one more than the other so don't mind what we go with but prob should pick one standard - do you have a pref?

tests/test_iasr_table_caching/test_local_cache.py
```diff
@@ -8,21 +8,30 @@ def test_build_required_tables_new_format():
         "ispypsa.iasr_table_caching.local_cache.FEATURE_FLAGS",
         {"use_new_table_format": True},
     ):
-        result = _build_required_tables()
-    assert result == [
+        result = _build_required_tables("7.5")
+    # Base topology tables come first
+    assert result[:4] == [
```
EllieKallmier: is the order of the tables important to preserve/test here?

EllieKallmier (approved):
Looking good! Left a few of the usual small questions/comments and have a few higher-level thoughts that I'll pop here to document (with no intent to block merging this in). 

Thought notes:
1. I think for me this PR is on the chunky side and possibly felt like could've been split up a tiny bit? Probably the only thing that feels like an easy split out is the table caching partitioning piece - and ofc I can see how interconnected it all is in the end so maybe there's not an obvious way to chunk this one. Either way not a big deal just noting for reflections' sake!
2. Sometimes I find the kind of chained/nested helper-function flow breaks up the description of a function's purpose/process a little obscured or not quick to grasp. E.g., the `_select_least_cost_option_per_expansion` function calls functions (that sometimes call other functions) - all good, makes sense - but from that function itself some of the detail of its process isn't obvious, like what "complete" costs means, or that the max. directional limit is used to determine the min cost/MW. I am certain that this is pure preference on my part here so I'm not suggesting we should repeat this kind of detail heaps through every docstring - I'm also thinking that this will prob be covered in the methodology docs rather than docstrings/API docs (?) just noticed myself sometimes needing to jump around a bit to see the full picture. 
3. Last thought/question - while reviewing this PR I was thinking about the role of the templater (again, lol) and just thinking that there's a bit of decision-making that happens here in terms of setting up the inputs. Mostly w.r.t the option choice/filtering step, where there's a bit of I guess loss of information when the least-cost option gets chosen? Knowing that we are also building in the regional granularity stuff that will also be a bit opinionated by design, I don't think this is wrong/bad - I'm not really sure what this thought is at the end of it lol more stream-of-consciousness sorry! 

Ok that's it from me, happy to go ahead with this bad boy :)

src/ispypsa/templater/transmission.py
nick-gorman: Good catch. This line was really just preventing a pandas warning when new_paths was empty. Which is better handled by fixing how _new_parallel_path_rows constructs the empty return case. I'll fix this.

src/ispypsa/templater/network_expansion.py
nick-gorman: Yeah, I think some defensive checks still make sense, particularly in the templater which is pre-any validation. You might be fine without them too because the IASR inputs are static and the checks just provide a bit of robustness to IASR inputs drift. Whether this one, which would only fire with zero flow path options, makes sense is arguable.

src/ispypsa/validation/schemas/network_expansion_options.yaml
nick-gorman: My bad, I'll change to the new format `- table_name: column_name`

tests/test_iasr_table_caching/test_local_cache.py
nick-gorman: No, that's actualy not great as it makes the test more fragile than it needs to be. I'll fix.

nick-gorman:
Thank @EllieKallmier, I apprecaite the stream-of-consciousness and all the comments. Very helpful to be aware of your takes.

(1) Noted. I'll keep looking to make the PRs more compact where possible.
(2) Interesting, I'll play around with the style, and let you know if I come up with anything / slash keep this in the back of my mind.
(3) Yep, totally, picking the least cost option is pretty arbitary and a bunch of info gets lost. The other option would be to keep all the options in network_transmission_path_expansion_costs, and then just specify which option is active in the network_expansion_options table (which could be least cost by default). Then users could just edit network_expansion_options to switch between options. Any thoughts on this approach?

EllieKallmier:
> Thank @EllieKallmier, I apprecaite the stream-of-consciousness and all the comments. Very helpful to be aware of your takes.
> 
> (1) Noted. I'll keep looking to make the PRs more compact where possible. (2) Interesting, I'll play around with the style, and let you know if I come up with anything / slash keep this in the back of my mind. (3) Yep, totally, picking the least cost option is pretty arbitary and a bunch of info gets lost. The other option would be to keep all the options in network_transmission_path_expansion_costs, and then just specify which option is active in the network_expansion_options table (which could be least cost by default). Then users could just edit network_expansion_options to switch between options. Any thoughts on this approach?

Sweet! Yeah cool, that seems like a nice straightforward op to handle this case - plenty of flexibility but a user could essentially ignore the network_transmission_path_expansion_costs 'extra' info without needing to interact/make decisions. Sounds good to me!

PR #107, opened 2026-05-25
Add new-format costs_connection templating (VRE new entrants)

EllieKallmier:
## Summary
1. Adds _template_connection_costs and supporting helpers in a new connection_and_build_costs.py module, templating the costs_connection table for VRE (wind and solar) new entrants in the new-format IASR workflow.
2. Adds _fuzzy_map_to_canonical (many-to-one canonical fuzzy matching) and two supporting helpers to helpers.py — set up for robust scenario name matching against raw IASR data, but hoping to apply to lots of cases as we progress
3. Updates _ISP_SCENARIOS to current IASR workbook scenario names (Slower Growth, Step Change, Accelerated Transition)
4. Small updates to costs_connection validation schema.
5. todo notes: some todo's left in here for me to apply when other changes/updates come through - e.g. #102 has some updates to structural stuff that I haven't touched in this PR.

#### Scope note:
Connection cost templating for non-VRE technologies is split out to come in a follow-up PR. The reasons:
- Non-VRE connection capacity data aren't directly reported in the IASR workbook - compiled across ENOR instead. So requires a bit of specific/fuzzy handling that I didn't want to chunk up this PR with.
- Batteries - being now defined for REZs as well just means the regional granularity stuff will also need to be handled kind of specifically, so thought easier to just split out into another PR.

The follow-up PR should build directly on this one. There are a few other things like more assertive canonicalisation to come as the rest of generator/storage templater updates built out, but for the sake of smaller chunks I'm leaning on the feature flag approach to more incrementally implement stuff - hope that's ok, and very happy to change that workflow/interpretation if it doesn't work long term!


tests/test_templater/test_create_ispypsa_inputs_template.py
```diff
@@ -167,6 +183,15 @@ def test_create_ispypsa_inputs_template_new_format_nem_regions(csv_str_to_df):
                 "renewable_energy_zones": renewable_energy_zones,
                 "flow_path_transfer_capability": flow_path_transfer_capability,
                 "initial_transmission_limits": initial_transmission_limits,
+                "connection_cost_forecast_wind_and_solar": csv_str_to_df(
+                    "REZ__ID, Scenario, 2024-25\nQ1, Step__Change, 73000000"
```
nick-gorman: Can put the data on a new line rather than using `\n` just a bit confusing visually

tests/test_templater/test_create_ispypsa_inputs_template.py
```diff
@@ -218,6 +256,15 @@ def test_create_ispypsa_inputs_template_new_format_single_region(csv_str_to_df):
                 "renewable_energy_zones": renewable_energy_zones,
                 "flow_path_transfer_capability": flow_path_transfer_capability,
                 "initial_transmission_limits": initial_transmission_limits,
+                "connection_cost_forecast_wind_and_solar": csv_str_to_df(
```
nick-gorman: Why no check for corresponding output?

src/ispypsa/templater/helpers.py
```diff
@@ -131,6 +131,79 @@ def _log_fuzzy_match(
...
+    """
+    canonical = set(choices)
+    match_dict = {
+        v: _best_fuzzy_match(v, canonical, threshold) for v in name_series.unique()
+    }
+    matched = name_series.map(
+        lambda v: match_dict[v] if match_dict[v] is not None else v
+    )
+    _log_fuzzy_match(name_series, matched, task_desc)
+    _warn_unmatched_fuzzy_values(name_series, match_dict, task_desc)
```
nick-gorman: Just a thought really. But if canonicalisation fails, this kind of implies that dataset produced will contain in consistencies and maybe shouldn't be used as is? Or maybe I'm missing something. But I guess validation would pick this up later.

src/ispypsa/templater/connection_and_build_costs.py
```diff
@@ -0,0 +1,573 @@
...
+    )
+    melted = melted.rename(columns=id_cols_rename)
+    melted = _enforce_numeric_cols(melted, ["connection_cost", "connection_capacity"])
+    melted["year"] = melted["year"].map(_financial_year_string_to_end_year_int)
+    return melted[
+        ["geo_id", "year", "connection_capacity", "connection_cost"]
+    ].reset_index(drop=True)
+
+
+def _enforce_numeric_cols(df: pd.DataFrame, numeric_cols: list[str]) -> pd.DataFrame:
```
nick-gorman, line 446: Good for me to remember this is here. Could be extracted to helpers down the track.

nick-gorman (approved):
Looks good Ellie. Some nice extensions to the fuzzy maping infrastructure. And very much apprecaite the chunking into a smallish PR! Just some tiny comments from me.

tests/test_templater/test_create_ispypsa_inputs_template.py
EllieKallmier: whoops yep not intentional! It'll return an empty costs df so I just blanked on checking that it actually exists. Will fix up :)

src/ispypsa/templater/helpers.py
EllieKallmier: yeah I think that's right, and I was thinking validation would pick it up but prob better to just raise as it happens (or drop?). The case that was in my mind is in the connection costs table for non-vre, 'BOTN - Cethana' has it's own entry in what is otherwise only standard 'technology'-labelled entries - so thought I'd just leave unmatched/unconverted so we could handle as a special case later. But I don't know how common that kind of case might be (and assume probably not very) so yeah probably better to handle known special cases before canonicalisation and just error out if canonicalisation fails. Will update!

Issue #110, opened 2026-05-28
PLEXOS battery inventory: what's shipped, buildable, and in ExportGroup constraints


nick-gorman:
## What this issue is

A side-by-side inventory of every battery PLEXOS ships in the
*Draft 2026 ISP Step Change Model.xml* against the batteries IASR ships
for the equivalent 7.5 workbook. The IASR side is drawn from two
workbook tables: `new_entrants_summary` (the 1h/2h/4h/8h + Distributed
Resources batteries) and `consumer_energy_resources_summary` (the
Coordinated CER Storages and V2G batteries, Fuel type == Battery).

Two tables follow — one for REZ-level batteries (keyed by REZ ID +
duration), one for sub-region-level batteries (keyed by sub-region +
duration). Each row shows the IASR name, the PLEXOS name, which IASR
table the match came from, the PLEXOS `Max Units Built` value, and
which `ExportGroup_*` constraints (if any) the PLEXOS object is a
member of.

The headline of why this exists: working through the custom-constraints
templater turned up several non-obvious choices in how PLEXOS represents
batteries (multiple LP variables per physical battery, most with
`Max Units Built = 0`; 4h variants present as objects but absent from
the export constraints; CER/V2G batteries that live in a separate IASR
table ISPyPSA doesn't currently template). This is a static record of
those facts that other workstreams can refer back to.

## Key findings

**LP variable counts:**

- PLEXOS ships **416 Battery objects** in total. **174** are existing /
  committed plants (or other non-new-entrant batteries) — these mostly
  have `Max Units Built` "Not set" or `0` because they're already built,
  not because they're non-buildable variants; they still carry their
  existing capacity.
- The remaining **242 are new-entrant battery LP variables** (REZ-named,
  sub-region-named, and constraint-scoped variants). **Only 39 of the
  242 have `Max Units Built > 0`** (can be built); 136 are `Max = 0`
  and 67 are "Not set". PLEXOS creates several parallel LP variables per
  physical new-entrant battery — e.g. a buildable sub-region-named
  variant plus a `Max = 0` constraint-scoped variant that sits inside
  the export constraint — so most new-entrant variants are LP variables
  that can never carry built capacity.

**REZ-named (`Q1 Battery - 2h` style) batteries:**

- IASR ships **3 batteries per REZ** (2h/4h/8h) across all 42 REZ IDs.
- PLEXOS ships matching objects for all but `V3` (no `V3 Battery - *`
[...]

PR #112, opened 2026-05-28
Add custom-constraints-from-PLEXOS templater

nick-gorman:
## What this does

ISPyPSA needs AEMO's REZ / sub-region export limits as custom constraints, but
they don't come out of the IASR workbook in a usable shape. AEMO does express
them in its PLEXOS model, as export-group constraints, so we extract them from
there. This PR adds the templater that turns that extract into ISPyPSA's three
custom-constraint tables:

- `custom_constraints`: one row per constraint (id + direction)
- `custom_constraints_lhs`: the left-hand-side terms (which unit, what coefficient)
- `custom_constraints_rhs`: the per-timeslice limit on the right

It sits behind the `use_new_table_format` flag (committed off) and only runs at
`sub_regions` granularity — coarser granularities collapse the sub-region
nodes, sub-regional flow paths and REZ-located units the constraints reference,
leaving nothing to constrain.

## What the translation looks like

One constraint end-to-end, based on the e2e test fixture. From the PLEXOS
extract (`lhs_terms.csv`, abbreviated — `constraints.csv` carries SWQLD1's
`Sense` of -1, `rhs_values.csv` its limit of 3000 MW tagged `"QLD Hot Day"`):

```
constraint_name     parent_class  parent_name                  property                 value
ExportGroup_SWQLD1  Generator     BW01                         Generation Sent Out C.    0.5
ExportGroup_SWQLD1  Generator     Q8_SAT_Brisbane              Generation Sent Out C.    0.4
ExportGroup_SWQLD1  Generator     SWQLD1_Linear Augmentation   Installed Capacity C.    -1.0  # relaxation term: dropped
ExportGroup_SWQLD1  Battery       SWQLD1 Battery - 2h          Generation Coefficient    1.0  # constraint-scoped: dropped, pass 2 re-supplies
ExportGroup_SWQLD1  Battery       SWQLD1 Battery - 2h          Load Coefficient         -1.0  # negative pair: dropped
ExportGroup_SWQLD1  Line          NSW1-QLD1                    Flow Coefficient          0.8
ExportGroup_SWQLD1  Purchaser     Q1 to NQ Flex. Electrolyser  Load Coefficient         -1.0  # hydrogen: dropped
```

the templater produces:

```
custom_constraints:
constraint_id  direction
SWQLD1         <=                                              # prefix stripped; Sense -1 -> <=

custom_constraints_lhs:
constraint_id  term_type         variable_name    coefficient  date_from
SWQLD1         generator_output  BW01             0.5
SWQLD1         generator_output  Q8_SAT_Brisbane  0.4
SWQLD1         link_flow         NSW-QLD          0.8          # PLEXOS line name -> path_id
SWQLD1         storage_output    Q8 Battery - 2h  1.0          # pass 2: Q8_SAT_Brisbane triggers REZ Q8
SWQLD1         storage_output    Q8 Battery - 4h  1.0          # incl. the 4h duration PLEXOS omits

custom_constraints_rhs:
constraint_id  timeslice        rhs     date_from
SWQLD1         qld_peak_demand  3000.0                         # "QLD Hot Day" -> region-prefixed timeslice
```

Note the RHS timeslice: it comes out region-prefixed (`qld_peak_demand`)
rather than bare (`peak_demand`) like the rest of the templater's outputs. A
follow-up PR will bring the rest of the transmission functionality onto this
format so the constraints flow through end-to-end.

## The bit worth understanding first: batteries

The RHS and most of the LHS are a straight translation from the PLEXOS
extract. The one judgement call is batteries.

PLEXOS encodes battery participation as per-constraint LP variables (the
`SWQLD1 Battery - 2h` rows in the example above) whose layout doesn't map onto
any IASR unit (see Open-ISP/ISPyPSA#110), and it quietly leaves out the 4h
durations. Rather than mirror that, the templater builds the LHS in two
passes:

1. **Translate.** Convert the raw PLEXOS LHS rows (generators, batteries,
   lines, nodes) into ISPyPSA's term schema. Unit names that don't resolve to
   an IASR unit are dropped — which removes the constraint-scoped battery
   variants by design.
2. **Re-inject.** For each constraint, find the REZs / sub-regions where its
   surviving new-entrant units (generators or batteries) sit, and add a
   `storage_output` term for every IASR new-entrant battery at those
   locations. Each injected battery copies the `(date_from, coefficient)`
   profile of the new-entrant batteries that survived pass 1 there, so an
   injected 4h battery lands at the same weight as its 2h/8h siblings (e.g.
   0.43 at SWQLD1's Q8) rather than a flat 1.0. Where no sibling survived, it
   falls back to 1.0 and a WARNING names the (constraint, location) so the
   default can be audited.

The copy is only well-defined if a location's surviving batteries agree, so
the templater raises if they don't share one profile.

More generally, the templater fails loud on anything it doesn't recognise: an
unmapped `(parent_class, property)` pair or `Sense` value, an unknown PLEXOS
line name, a battery `Load Coefficient` row missing its `Generation
Coefficient` pair, or a populated `date_to`. Deliberate drops are logged —
INFO where they're by-design (Purchaser electrolysers, constraint-relaxation
terms, battery Load pairs), WARNING where the drop loosens a constraint (units
with no ISPyPSA counterpart).

## Where it lives

```
ISPyPSA/
├─ src/ispypsa/templater/
│  ├─ custom_constraints_from_plexos.py   the templater (PLEXOS extract → the 3 tables)
│  ├─ plexos/7.5/*.csv                    the extract it reads at runtime (already on main)
│  ├─ create_template.py                  wires it in at sub_regions; tracks the 3 outputs
│  └─ mappings.py, transmission.py        now share one canonical-timeslice vocabulary
├─ scripts/
│  ├─ extract_plexos_constraints.py       PLEXOS XML → the extract above (already on main)
│  └─ workbook_extract_constraint.py      independent workbook-route LHS, validation only
└─ tests/test_templater/
   ├─ test_custom_constraints_from_plexos.py   templater unit tests
   ├─ test_custom_constraints_validation.py    PLEXOS route vs workbook route reconciliation
   └─ data/custom_constraints_validation/      fixtures (plexos/ + workbook/)
```

## Supporting changes

The templater is wired into `create_ispypsa_inputs_template` (gated to
`sub_regions`, its three outputs tracked as task targets); the two IASR summary
tables it needs are added to the required-tables list; and the canonical
timeslice vocabulary is promoted into `mappings` so this templater and the
existing transmission code share one copy.

## Validation

Pulling these constraints out of PLEXOS takes enough translation that we
didn't want to take the output on trust, so the PR cross-checks the templater
against a second, fully independent reconstruction of each constraint built
straight from the IASR workbook. The two routes share no code:

- The **PLEXOS route** is the production templater reading the PLEXOS extract.
- The **workbook route** (`scripts/workbook_extract_constraint.py`) looks each
  constraint up in `rez_group_constraint_summary`, parses its Terms into
  `(coefficient, body)` pairs, and expands them: a REZ id (e.g. `Q1`) becomes
  every generator, battery and electrolyser tagged with that REZ ID across the
  IASR unit summary tables; an interconnector path id (e.g. `CQ-NQ`) becomes a
  single `link_flow` term. Coefficients can be implicit (`Q1` is 1.0), signed
  (`- CQ-NQ` is -1.0) or explicit (`0.78 * V7`).

`test_custom_constraints_validation.py` runs both routes over five constraints
(`NQ1`, `NET1`, `WV1`, `CQ1`, `MN1`) and asserts the LHS term sets match
exactly, save for an explicit `EXPECTED_DELTAS` table. A new mismatch in
either direction fails the test, and so does an expected one going missing.

The documented differences come down to two things:

1. **Geographic lookup for existing plant.** Every IASR unit carries both a
   fine-grained REZ ID and a coarser sub-region, and they usually agree. The
   workbook route expands by REZ ID; PLEXOS includes existing plant by
   sub-region. The handful of boundary plants where the two tags point
   different places are the only generator diffs — e.g. `EMERASF1` (REZ `Q4`,
   sub-region `NQ`) lands in `NQ1` under PLEXOS but `CQ1` under the workbook,
   so the one fact shows up on both sides of the comparison.
2. **Electrolysers.** PLEXOS routes electrolyser load through its `Purchaser`
   class, which the templater drops by design (hydrogen demand isn't
   modelled), so they appear as workbook-only `load` terms.

Batteries don't surface as a difference: the five validated constraints are
all unit-coefficient, so both routes land the same batteries at 1.0. The
pass-2 coefficient copy only bites on the fractional constraints (SWQLD1,
WNV1, NSA1, SQ1), which the workbook route can't reach — those rely on the
per-module unit tests instead.


scripts/workbook_extract_constraint.py
```diff
@@ -0,0 +1,277 @@
+"""Manually extract a rez-group custom constraint from the IASR workbook.
+
+This is an independent reconstruction of one custom constraint, built directly
```
EllieKallmier, line 3: To clarify/summarise - this is a helper script whose output is just to print the LHS unit rows processed for the given constraint ID?

src/ispypsa/templater/create_template.py
```diff
@@ -83,12 +86,25 @@
     "network_transmission_path_expansion_costs",
 ]
 
+# Custom constraints are templated only at sub_regions granularity (see the gate
+# in create_ispypsa_inputs_template) — coarser granularities collapse the
+# sub-region nodes, sub-regional flow paths and REZ-located units the constraints
+# reference, leaving nothing to constrain. Listed as outputs only at that
+# granularity so the create_ispypsa_inputs task tracks them where they are
+# written and does not expect them where they never are.
```
EllieKallmier, line 94: This I guess supports an approach to regional granularity (for the moment) where something stops users from actually running any granularity other than subregional (e.g. feature flags). 

Also was just having a look back at [this issue](https://github.com/Open-ISP/dev-practices/issues/7) re: naming conventions and I think "subregions" (no hyphen/underscore) was the decision there for node naming - absolutely the most minor comment and I have not been paying attention to this much myself but just chucking here in case we want to chat/change/do anything in relation to how we refer to sub(-)regions in other contexts?

src/ispypsa/templater/custom_constraints_from_plexos.py
```diff
@@ -0,0 +1,1172 @@
...
+
+import pandas as pd
+
+from .mappings import _CANONICAL_TIMESLICES
+
+# PLEXOS REZ-id prefixes that IASR renamed to ``DREZ``. Applied to the first
+# token of generator names (``DN1_SAT_Dubbo`` -> ``DREZ_SAT_Dubbo``) and to
+# battery names of the form ``"DN1 <location> Battery - <duration>"``
+# (``DN1 Dubbo Battery - 2h`` -> ``DREZ Dubbo Battery - 2h``).
+_GENERATOR_PREFIX_RENAME = {"DN1": "DREZ", "DN3": "DREZ"}
```
EllieKallmier, line 184: Here and for batteries below - why no DN2? (Guessing this DREZ isn't present in the plexos constraints?)

src/ispypsa/templater/custom_constraints_from_plexos.py
EllieKallmier, line 1: I think this file is where that thing you mentioned in the meeting last week about phrasing getting a little confusing/maybe odd is coming through for me. Some of the words ('trigger', 'surviving') do make sense and aren't _not_ understandable, but maybe could be introduced or defined a bit more specifically before going on to be used kind of ubiquitously? 

Just feeling a bit like sometimes adding a few extra words or something could make some of the docstrings a little easier to intuit - but of course this is a personal take and appreciate is a bit in the weeds! 

To give an example, this sentence from one of the docstrings: 

> Surviving batteries trigger as well as generators so a location
> whose generators were all dropped (or that PLEXOS includes via batteries
> alone) still gets its omitted durations re-added.

is kind of isolated from previous introduction of the 'surviving' and 'trigger' terms; adding a few extra words to contextualise is prob my instinct from a clarity POV, e.g. "Surviving batteries trigger the creation of additional rows even in locations whose generators were all dropped ...". 

This is not a big deal and I was able to wrap my head around this file in the end but yeah I think the use of some of these terms or phrases that have a kind of implied context-specific meaning just added some extra time for me in unpicking it. Again - prob a person-by-person thing so no change required from me, just feeding back on what we chatted about the other day!

src/ispypsa/templater/custom_constraints_from_plexos.py
```diff
@@ -0,0 +1,1172 @@
...
+            f"Applied {len(pairs)} PLEXOS->ISPyPSA LHS name renames: {renames}"
+        )
+
+
+def _drop_unresolved_terms(lhs: pd.DataFrame) -> pd.DataFrame:
+    """Drop LHS terms whose parent_name resolved to no ISPyPSA unit.
+
+    These are PLEXOS units with no ISPyPSA counterpart -- e.g. CER units not
+    yet modelled (Open-ISP/ISPyPSA#104) or plant absent from the IASR tables.
+    Dropping loosens the constraint, so it is logged as a WARNING.
```
EllieKallmier, line 735: No real comment just some out-loud thinking for me to reflect on:

Makes sense to drop and log as warning here, and/but kinda relates to the canonicalisation question you raised in #107  and our discussions around what we expect the templater input to look like. Basically I'm wondering if (as you posited in a comment on that pr) it might make more sense or be simpler to just drop/warn/not raise on non-matches and let the validator raise 'real' problems later (i.e. not just 'not implemented yet' kinda things)... in some cases

src/ispypsa/templater/custom_constraints_from_plexos.py
```diff
@@ -0,0 +1,1172 @@
...
+            SWQLD1         storage_output  Q8 Battery - 8h  0.43
+            WNV1           storage_output  V7 Battery - 2h  0.78
+            WNV1           storage_output  V7 Battery - 2h  0.00         2031-11-30
+            SWQLD1         storage_output  Tarong BESS      0.14         # existing -> no location, ignored
+
+        returns:
+            constraint_id  location  coefficient  date_from
+            SWQLD1         Q8        0.43
+            WNV1           V7        0.78
+            WNV1           V7        0.00         2031-11-30
```
EllieKallmier: I am getting a little bit mixed up with some of the wording here and what the ``_raise_on_inconsistent_battery_profiles`` function below actually checks for 'inconsistency' - very possible I'm missing something about constraint definitions so grain of salt!

Specifically: 
- "the surviving new-entrant batteries **must share one** (date_from, coefficient) profile" -> reads to me like saying **only one** (date_from, coefficient) tuple is allowed for each (constraint, location) tuple - in which case I think I assume that the I/O example should raise for the (WNV1, V7) pair? 
- "a time-varying sibling whose extra date_from rows would otherwise be grafted onto its constant siblings silently" (below) -> describes a failure model -> to me reads like it's saying only either time varying or constant (but not both) pairs allowed, which again the I/O example here would raise (?)

From the tests and the code itself I can see this is not the correct interpretation - think maybe these two docstrings and/or example here could be edited a little to improve clarity?

tests/test_templater/test_custom_constraints_from_plexos.py
```diff
@@ -0,0 +1,1485 @@
...
+        constraint_id,  location,  coefficient,  date_from
+        SWQLD1,         Q8,        0.43,
+        WNV1,           V7,        0.78,
+        WNV1,           V7,        0.0,          2031-11-30T00:00:00
+    """)
+
+    result = _battery_rows_for_triggers(triggered, batteries_by_location, coefficients)
+
+    # Both Q8 batteries inherit 0.43; the V7 battery is emitted once per profile
+    # row (0.78 always, 0.0 from 2031-11-30); Q1's battery defaults to 1.0.
```
EllieKallmier: I'm a bit confused by the phrasing "0.78 always, 0.0 from date" - possibly just not understanding constraint formulation but "always" then "from x" seem contradictory? Or jsut not very clear - like "0.78 up to date, 0.0 from date" would be more obvious to me (related to earlier comment/confusion around this function's definition) - but perhaps I've also got an incorrect interpretation of what these values mean

tests/test_templater/test_custom_constraints_from_plexos.py
```diff
@@ -0,0 +1,1485 @@
...
+        IASR ID / DLT names,  Sub-region,  REZ ID,  Technology Type
+        Q8 Battery - 2h,      SQ,          Q8,      Battery Storage (2hrs storage)
+        Q8 Battery - 8h,      SQ,          Q8,      Battery Storage (8hrs storage)
+    """)
+
+    with pytest.raises(ValueError, match="share one"):
+        _surviving_battery_coefficients(lhs, new_entrants)
+
+
+def test_surviving_battery_coefficients_raises_on_divergent_date_sets(csv_str_to_df):
```
EllieKallmier, line 1342: Ok I think this explains this function a bunch more for me - I understand the vibe here! 

Still though think there could be some clarification within the docstrings/examples to make this a bit more intuitive or obvious there?

EllieKallmier (approved):
I think this looks good! My main comments/thoughts on this PR are mostly around my own confusion with some of the descriptions of stuff and phrasing used - hopefully my comments below clarify but ofc happy to discuss more. 

The two-pass approach makes sense, and matches up with behaviour elsewhere re: essentially merging in the 'missing' rows where there's some indication that they should exist (and no indication that they explicitly shouldn't). 

I don't think any of my comments should stop this PR from merging - I'm hoping I've understood at least the main vibes of what's going on and based on that understanding this work seems solid :)

scripts/workbook_extract_constraint.py
nick-gorman: Yes, the out puts are then used to help validate the constraints being produced by the Plexos path here: tests/test_templater/test_custom_constraints_validation.py

src/ispypsa/templater/create_template.py
nick-gorman: Yep, It'd been in the back of my mind to do a single PR that switches to the new subregions covention. But I hadn't bother writing new stuff to that convention untill the change of PR happens.

src/ispypsa/templater/custom_constraints_from_plexos.py
nick-gorman: Yep, DN2 doesn't appear in the plexos constraints data. Might follow up with a quick check that this makes sense that its missing.

tests/test_templater/test_custom_constraints_from_plexos.py
nick-gorman: Yeah, that was just badly worded now it reads:

```
    # Both Q8 batteries inherit 0.43; the V7 battery is emitted once per profile
    # row (0.78 until 2031-11-30, then 0.0 from 2031-11-30); Q1's defaults to 1.0.
```

tests/test_templater/test_custom_constraints_from_plexos.py
nick-gorman: I'm updating the module level docstring which will hopefully set up the explination of coeffients profiles much better.

src/ispypsa/templater/custom_constraints_from_plexos.py
nick-gorman: Great point. Commit [7024be3](https://github.com/Open-ISP/ISPyPSA/commit/7024be318f4d148fa2f83a9f3a87126e6f15606c) attempts to address this and I think does a reasonable job. Note it also establishes a guide in the CLAUDE.md to try and fix this type of issue going forward. However, as always I intend the CLAUDE.md as a work in progress (trying things to see how they work and if we like them), so we can continue to refine.

PR #114, opened 2026-06-09
Template new non vre connection costs

EllieKallmier:
Add templating functionality (and tests) for non-VRE new entrant connection costs. Includes:
- Created manual table ``connection_capacity_non_vre`` -> currently saved in ``manually_extracted_template_tables/7.5`` but open to moving elsewhere as its function is slightly different to other manual tables. To manage this I've just manually popped the read-in df out of manual table dict and into IASR table dict - again open to suggestions/other preferences for handling
- Regional granularity processing/filtering applied - non-VRE connection costs are given by technology and NEM region. Rows added for non-VRE defined in REZs (2, 4, 8hr batteries only atm)
- Some tweaks to overall flow that also touch VRE connection costs - tests and functions updated accordingly (primarily: no longer passing full ``*_new_entrant`` tables to the orchestrator functions, rather pre-processing to only pass necessary data through)

There are still a few notes that point to open discussions or might need addressing but at this point felt non-urgent and easier to come back to once we are a few steps further into validation and consensus with #103 . 

If I can also request in particular feedback on some of the tests - I don't know if I've done a great job on covering all necessary cases without too much double up so if anything sticks out as particularly unnecessary/messy or could be handled more clearly please let me know and I'll fix :)

### DIFF
--- a/src/ispypsa/templater/connection_and_build_costs.py
+++ b/src/ispypsa/templater/connection_and_build_costs.py
@@ -76,7 +76,7 @@ def _template_connection_costs(
     Returns:
         One row per (geo_id, technology, year). connection_cost and
         system_strength_cost are in $/MW; system_strength_cost is 0.0 for
-        non-IBR technologies and offshore wind.
+        non-IBR technologies and solar thermal.
 
     I/O Example:
         Inputs (abbreviated; regional_granularity="nem_regions"):
--- a/src/ispypsa/templater/connection_and_build_costs.py
+++ b/src/ispypsa/templater/connection_and_build_costs.py
@@ -19,113 +19,146 @@ _VRE_COLUMN_RENAMES = {
     "REZ ID": "geo_id",
     "Connection capacity (MVA)": "connection_capacity",
 }
+_NON_VRE_COLUMN_RENAMES = {
+    "Region": "region_id",
+    "Generator Type": "technology",
+    "Connection capacity (MVA)": "connection_capacity",
+}
 
+# NOTE: the VRE / non-VRE split below is approximate, not strict. The IASR
+# "connection_cost_forecast_other" (non-VRE) table actually includes "distributed
+# solar", so "non-VRE" really means "everything templated outside the wind/solar
+# REZ-connection path". This is a known wart — revisit how distributed resources
+# are handled once more of the templater (and schema validation) is in place.
 _VRE_TECHNOLOGY_STRINGS = ["solar", "wind"]
 _IBR_TECHNOLOGY_STRINGS = _VRE_TECHNOLOGY_STRINGS + ["battery", "batteries"]
+_NON_VRE_EXCLUDED_TECHNOLOGIES = ["Alkaline Electrolyser", "BOTN - Cethana"]
 
 
 def _template_connection_costs(
-    connection_cost_forecast_vre: pd.DataFrame,
-    connection_costs_for_vre: pd.DataFrame,
-    system_strength_cost_table: pd.DataFrame,
+    iasr_tables: dict[str, pd.DataFrame],
     scenario: str,
+    regional_granularity: str,
     generators_new_entrant: pd.DataFrame,
+    storage_new_entrant: pd.DataFrame,
+    sub_regional_geography: pd.DataFrame,
 ) -> pd.DataFrame:
     """Combines VRE and non-VRE connection costs with system strength costs.
 
     Produces a long-format table with columns (geo_id, technology, year,
-    connection_cost, system_strength_cost) from three IASR source tables.
-    VRE (wind/solar) costs are templated and system strength costs are merged
-    in to return a single ISPyPSA format table: ``costs_connection``.
-
-    Note: non-VRE connection costs will be added in a subsequent PR.
+    connection_cost, system_strength_cost). VRE (wind/solar) costs are templated
+    per-REZ; non-VRE costs are templated per-region and expanded to geo_id level
+    based on ``regional_granularity``. System strength costs ($/MW) are merged in
+    for all IBR technologies (wind, solar, battery).
 
     Args:
-        connection_cost_forecast_vre: IASR ``connection_cost_forecast_wind_and_solar``
-            table. Columns: 'REZ ID', 'Scenario', one column per financial year.
-        connection_costs_for_vre: IASR ``connection_costs_for_wind_and_solar`` table.
-            Columns: 'REZ ID', 'Connection capacity (MVA)'.
-        system_strength_cost_table: IASR ``efficient_level_of_system_strength_cost``
-            table. Single row with one column per financial year (cost in $/kW).
+        iasr_tables: dict containing the following keys:
+            ``connection_cost_forecast_wind_and_solar``: columns REZ ID, Scenario,
+                one column per financial year (total connection cost in $).
+            ``connection_costs_for_wind_and_solar``: columns REZ ID, Connection
+                capacity (MVA).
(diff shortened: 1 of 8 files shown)


src/ispypsa/templater/create_template.py
```diff
@@ -203,13 +205,23 @@ def create_ispypsa_inputs_template(
         # todo: replace with actual generators_new_entrant once that templating
         # function is written — passing empty placeholder for now so costs_connection
         # is wired up but produces no VRE rows until generators are templated.
+
+        # connection_capacity_non_vre is in manually_extracted_template_tables/ (sourced from
+        # ENOR tables 16-17 and confirmed with AEMO) but is needed as an iasr_tables input,
+        # not a template output.
+        iasr_tables["connection_capacity_non_vre"] = manually_extracted_tables.pop(
```
nick-gorman: In PR112 I've changed the templater to read the manul table directly from disk rather than recieve them as an input. I just thought it was a bit simpler (or hides the complexity). I'm happy to go either way, what do you think?

Also, if its coming as an input, maybe we shouldn't use pop, as this modifies somthing the user parsed in, and I guess its possible they might still want connection_capacity_non_vre in manually_extracted_tables.

src/ispypsa/templater/connection_and_build_costs.py
```diff
@@ -19,113 +19,146 @@
...
+            columns used: 'geo_id' and 'technology'.
+        sub_regional_geography: templated network geography table;
+            columns used: 'geo_id', 'geo_type', 'region_id'.
 
     Returns:
         One row per (geo_id, technology, year). connection_cost and
-        system_strength_cost are in $/MW; either may be NaN where no cost
-        applies for that combination.
+        system_strength_cost are in $/MW; system_strength_cost is 0.0 for
+        non-IBR technologies and offshore wind.
```
nick-gorman: Should there be an explicit setting of offshore wind to zero as for non-IBR (_set_non_ibr_system_strength_cost_to_zero and _set_solar_thermal_system_strength_costs_to_zero)?

src/ispypsa/templater/connection_and_build_costs.py
```diff
@@ -286,6 +316,362 @@ def _build_vre_cost_rows(
...
+    )
+    normalised_df = _normalise_connection_cost_forecast_frame(
+        merged_df, id_cols_rename=_NON_VRE_COLUMN_RENAMES
+    )
+    canonical_technologies = set(canon_tech_geo_id_pairs["technology"])
+    canonicalised_df = _canonicalise_non_vre_technologies(
+        normalised_df, canonical_technologies
+    )
+    costs_per_mw = _calculate_connection_cost_per_mw(canonicalised_df)
+    _warn_nan_connection_costs(costs_per_mw, id_cols=["region_id", "technology"])
```
nick-gorman, line 405: Is this currently firing on the 2026 input data? Curious what to make of the MW what values being zero.

src/ispypsa/templater/connection_and_build_costs.py
nick-gorman: Oh, I think I see now, this is a outcome of there not being any forecast costs for offshore wind.

tests/test_templater/test_create_ispypsa_inputs_template.py
```diff
@@ -144,15 +144,22 @@ def test_create_ispypsa_inputs_template_new_format(csv_str_to_df):
         REZ / Constraint ID,  Option,    2024-25,  2025-26
         N3,                   Option 1,  750000,   760000
     """)
-
     connection_cost_forecast_wind_and_solar = csv_str_to_df("""
-        REZ__ID,  Scenario,     2024-25
-        Q1,       Step__Change, 73000000
+        REZ__ID,  Scenario,     2024-25,    2025-26
```
nick-gorman: I don't think we need the `__` anymore, and csv_str_to_df should be able to handle just normal spaces.

tests/test_templater/test_connection_and_build_costs.py
```diff
@@ -527,7 +476,7 @@ def test_merge_and_filter_system_strength_costs(csv_str_to_df):
 def test_merge_and_filter_system_strength_costs_year_not_in_system_strength_gives_nan(
```
nick-gorman, line 476: Not really a comment on this PR. But just saw this when reading through the test file. It's an interesting case, which might happen in many places. I guess if we wanted to inforce no missing years then the validator would be the place to do this. So yep, I think filling with nan makes sense.

tests/test_templater/test_connection_and_build_costs.py
```diff
@@ -861,61 +817,635 @@ def test_template_vre_connection_costs_empty_capacity_gives_nan_costs(csv_str_to
...
+    costs = csv_str_to_df("""
+        region_id,  technology,             year,   connection_cost
+        SA,         Battery__Storage__(4h), 2025,   90000.0
+    """)
+    empty_pairs = pd.DataFrame(columns=pairs.columns)
+    empty_geography = pd.DataFrame(columns=geography.columns)
+    empty_costs = pd.DataFrame(columns=costs.columns)
+
+    expected = pd.DataFrame(columns=_CONNECTION_ONLY_COST_COLS)
+    # A emtpy
```
nick-gorman: typo in a few places

nick-gorman (approved):
Hey, Ellie this is looking good. A few minor comments. I left my main substaintal comment as a discussion post (#115) because I think it's broader than this PR. However, I think the point raised there would be the biggest thing to reduce test duplication, I'm keen to hear your thoughts on it.

src/ispypsa/templater/connection_and_build_costs.py
EllieKallmier: yeahh this is probably a "too flexible" moment tbh - I might go back and simplify some bits here after our chat this morning. There are some rows in the 'capacity' df for VRE that have NaNs, so I thought I'd just cover the divide by 0 case just in case it popped up also - but post-chat this morning that is prob a bit extra and overcomplicating things :)

src/ispypsa/templater/create_template.py
EllieKallmier: Yeah that sounds fine to me I think! I've taken a look at that PR and can't immediately spot what you're referring to but by this description I'm on board hahah

Re: popping, yeah I didn't love mutating the manual dict so open to not doing that - was just looking at the old version of things and didn't want this table to end up as a templater output, but ofc we can just handle that differently in the new version lol :) will update!

src/ispypsa/templater/connection_and_build_costs.py
EllieKallmier: yeah, that was the goal I think - but very happy to make explicit too for clarity/self docs purposes.

src/ispypsa/templater/create_template.py
nick-gorman: Hey, pardon the confusion for my comment. I was refering to the way the plexos tables were read from disk within create_ispypsa_inputs_template rather than provided as inputs. So at the moment its a bit split up with plexos and manual IASR tables treated differently. I guess ideally they follow the same pattern, but now I'm unsure myself what is best haha

src/ispypsa/templater/connection_and_build_costs.py
EllieKallmier: actually no - this is my bad memory and an unfixed docstring, thanks for catching lol! Offshore wind gens do incur an additional system strength remediation cost, at least according to the workbook new entrants summary table, they just don't have an additional connection cost (that's included in the build cost). Idk why that's how it is but who am I to know these thigns

Issue #109, opened 2026-05-27
Network templater should map IASR timeslices to region-tagged PLEXOS timeslices using destination/origin rule


nick-gorman:
## Background

The IASR-derived transmission templater (`src/ispypsa/templater/transmission.py`) produces region-agnostic timeslices on flow-path limits — `peak_demand`, `summer_typical`, `winter_reference` — paired with a `direction` of `forward` / `reverse`.

The PLEXOS model, by contrast, attaches **region-prefixed** timeslice tags to `Line` properties: `"NSW Hot Day"`, `"QLD Typical Summer"`, `"VIC Winter"`, etc. For intra-region paths the right region is obvious, but for inter-regional paths the mapping isn't.

## What the PLEXOS model actually does

Investigation in `notes/line_timeslice_poke.py` (walks every `Line` object in the AEMO Draft 2026 ISP Step Change PLEXOS XML and tabulates the tag-region on `Max Flow` vs `Min Flow`) found a clean, zero-exception rule:

- **`Max Flow`** (forward, positive limit) carries the **destination** region's timeslice tag.
- **`Min Flow`** (reverse, negative limit) carries the **origin** region's timeslice tag.

This makes physical sense: the seasonal demand condition that constrains the limit is in the receiving region — that's where load tightens the import limit. The rule is symmetric, so reversing direction flips the tag region.

The same rule holds for `Max Capacity Reserves` / `Min Capacity Reserves`. Intra-region paths collapse to the single shared region, which is what the existing code already implicitly assumes.

### Per-line evidence (all 18 Lines in the AEMO model)

| Line | from -> to | MaxFlow tag region | MinFlow tag region |
|------|------------|--------------------|---------------------|
| EnergyConnect | SNSW -> CSA | SA | NSW |
| Marinus | TAS -> SEV | VIC | TAS |
| N-Q-MNSP1 | NNSW -> SQ | QLD | NSW |
| NSW1-QLD1 | NNSW -> SQ | QLD | NSW |
| T-V-MNSP1 | TAS -> SEV | VIC | TAS |
| VIC1-NSW1 | WNV -> SNSW | NSW | VIC |
| V-SA | WNV -> SESA | SA | VIC |
| V-S-MNSP1 | WNV -> CSA | SA | VIC |
| *(intra-region lines)* | ... | (single region) | (same region) |

## Required change

[...]

PR #116, opened 2026-06-10
Bring network and resource schemas up to the newer schema format

nick-gorman:
The costs, generators and storage schemas were restructured in #93 with sectioned descriptions (`Source tables` / `Source notes` / `If absent`), per-column absence semantics and numeric value constraints, but the earlier `network_` and `resource_` schemas kept the old flat style. This aligns them so all schemas share one format ahead of the validation layer landing. Stacked on #109.

```
├── network_geography.yaml                        ← source tables; subregion_id absence semantics
├── network_transmission_paths.yaml               ← source tables; parallel-corridor source note
├── network_transmission_path_limits.yaml         ← blank capacity/timeslice meanings; gte on capacity
├── network_expansion_options.yaml                ← least-cost selection source note; gte on expansion
├── network_transmission_path_expansion_costs.yaml ← integer → int; gte on cost
└── resource_limits.yaml                          ← source table; hard-cap penalty semantics
```

Two choices a reviewer might query:

- **No `nan_fill` anywhere.** In these tables a blank cell carries meaning (no physical limit, hard cap, static timeslice) rather than standing in for a default value.
- **`resource_limits` source table is inferred.** It has no new-format templater yet, so `initial_build_limits` comes from the legacy REZ build-limits templater and may need revisiting when the real one lands.

The `gte: 0.0` constraints are new spec, not just reformatting — nothing enforces them today, but they become real rules once validation lands.

### DIFF
--- a/src/ispypsa/validation/schemas/network_transmission_path_expansion_costs.yaml
+++ b/src/ispypsa/validation/schemas/network_transmission_path_expansion_costs.yaml
@@ -31,7 +31,7 @@ columns:
   year:
     type: int
     required: true
-    description: Financial year the cost applies to.
+    description: Year the cost applies to.
   cost:
     type: float
     required: true
--- a/src/ispypsa/validation/schemas/network_transmission_path_expansion_costs.yaml
+++ b/src/ispypsa/validation/schemas/network_transmission_path_expansion_costs.yaml
@@ -3,11 +3,24 @@ required: false
 unique:
   - [expansion_id, year]
 description: >
-  Time-varying annualised expansion costs for network elements, covering both
-  physical transmission paths and group constraints.
-  Merges flow path and REZ transmission expansion costs into a single long-format
-  table.
-  If this table is absent, network expansion is treated as free.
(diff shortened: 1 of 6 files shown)


src/ispypsa/validation/schemas/network_transmission_path_expansion_costs.yaml
```diff
@@ -16,13 +29,14 @@ columns:
       - network_expansion_options: expansion_id
     description: Identifier for the expandable network element.
   year:
-    type: integer
+    type: int
     required: true
     description: Financial year the cost applies to.
```
EllieKallmier: Minor note I think you pointed out earlier - we could just use "Year" rather than "Financial year" in these descriptions to be a little freer (unless is distinct here?)

EllieKallmier (approved):

src/ispypsa/validation/schemas/network_transmission_path_expansion_costs.yaml
nick-gorman: Yes! nice catch!

PR #124, opened 2026-06-19
Template new format new entrants identity cols

EllieKallmier:
This PR is the first of a few to come in the process of templating new entrants - both generators and storage. 

Scope: 
- initial templating for 'identity' columns only for generators_new_entrant and storage_new_entrant
- wiring into `create_ispypsa_inputs_template` to feed into connection costs templating
- minor update to storage_new_entrants schema (removing 'Power Station' column)

Coming up in next PR(s):
- Writing new-format templated new_entrant tables to templater output
- Filling out extra columns!

For the moment I've set up both new entrant groups to live in the same file, but this might change/split in future as more detail gets added to each - and as existing/planned technologies get templated too. Happy to take on any suggestions around this kind of structural stuff if there are any :)


src/ispypsa/templater/new_entrants.py
```diff
@@ -0,0 +1,237 @@
...
+    "WH": "wind_high",
+    "WM": "wind_medium",
+    "WFX": "wind_offshore_fixed",
+    "WFL": "wind_offshore_floating",
+    "SAT": "solar",
+    "CST": "solar",
+}
+
+# Extraction pattern for the resource-quality code embedded between underscores in
+# a VRE IASR ID, e.g. "WFX" in "N10_WFX_Hunter Coast".
```
nick-gorman:
```suggestion
# Regex extracting the resource-quality code embedded between underscores in a VRE                                                                                                                                                  # IASR ID, e.g. "WFX" in "N10_WFX_Hunter Coast". Derived from the code map, it                                                                                                                                                    
# expands to "_(WFX|WFL|SAT|...)_" — one capture group over the known codes,                                                                                                                                                      # sorted longest-first so a short code can't shadow a longer one it prefixes.
```

tests/test_templater/test_new_entrants.py
```diff
@@ -0,0 +1,200 @@
...
+    expected_storage = csv_str_to_df("""
+        Technology Type,                    REZ ID
+        Battery Storage (2hrs storage),     N3
+        Distributed Resources Batteries,    Not Applicable
+        Pumped Hydro (24hrs storage),       Not Applicable
+    """)
+    pd.testing.assert_frame_equal(storage, expected_storage)
+
+
+def test_filter_to_technology_group_raises_unknown_group(csv_str_to_df, caplog):
```
nick-gorman: Nit picking / totally fine the way it is. But since _filter_to_technology_group is a private function (and you know it will only be fed storage or generators) you could probably not have the raise and assocaited tests.

src/ispypsa/templater/new_entrants.py
```diff
@@ -0,0 +1,237 @@
...
+        return new_entrants_summary.loc[is_storage].reset_index(drop=True)
+    if group == "generators":
+        return new_entrants_summary.loc[~is_storage].reset_index(drop=True)
+    raise ValueError(
+        "Filtering new entrants table to technology group: "
+        f"group must be 'generators' or 'storage', got {group!r}"
+    )
+
+
+def _set_geo_id(new_entrants: pd.DataFrame) -> pd.DataFrame:
```
nick-gorman, line 132: Somewhat duplicates custom_constraints_from_plexos.py:938 _pick_location. We could extract to a helper, but it doesn't concern me too much,

src/ispypsa/templater/new_entrants.py
```diff
@@ -0,0 +1,237 @@
...
+def _filter_to_technology_group(
+    new_entrants_summary: pd.DataFrame, group: str
+) -> pd.DataFrame:
+    """Returns the summary rows for one technology group: generators or storage.
+
+    Storage rows are those whose "Technology Type" contains a
+    ``_STORAGE_TECHNOLOGY_STRINGS`` substring (battery, pumped hydro), matched
+    case-insensitively; generators are every other row. The two groups partition
+    the summary, so this single predicate is the only place the generator/storage
+    boundary is defined.
```
nick-gorman: custom_constraints_from_plexos.py:935 _is_battery_row ("Batter"); and                 static_new_generator_properties.py:65 (r"battery|pumped hydro") are also doing this work. Not suggesting we extract helpers (I don't think). But maybe loosen the language in the comment?

src/ispypsa/templater/new_entrants.py
nick-gorman: Could also guard against future REZ ID being NaN rather than "Not Applicable", but that goes against our lets just assume the templater inputs are fixed principal, on the other hand its a cheap bit of robustness.

nick-gorman (approved):
I think the structure is good, very readable, and a nice size for review. Minor comments only.

nick-gorman:
Actually, on thing that might have slipped past me in the review (and pervious ones) is adding new format templater tables to this CLI test test_create_ispypsa_inputs_new_table_formats.py::test_create_ispypsa_inputs_new_format, which runs real IASR data through the templater using the CLI and then checks the results are roughly the shape we expect. Not a blocker, and something it might be easier to circle back to when the new templater is being finalised.

EllieKallmier:
Some notes from post-review updates:
- I did end up pulling out some helpers in places you highlighted, as I do think they made sense and let me cut down a bit particularly in the tests. As part of this I added an equivalent pumped hydro helper to match the _is_battery_row one and a little wrapper to return their union; these are only used in the one spot at the moment but I can foresee a few spots where I think having these extras will come in handy
- I also though did leave a (possibly unnecessary?) essentially wrapper dude in the new_entrants file pretty much because it was prettier to keep (_set_geo_id) and keeps a nicer readable set of steps in the orchestrators

Re: your comment about the CLI tests - yep good point, and I think I will circle back after I add the next prs with more detail onto this one and actually wire the new_entrants outputs to be templater outputs as well. So I'll do a check + add for everything that I've missed that on so far at that point if ok :)

PR #125, opened 2026-06-24
Make the new-format templater output set granularity-invariant

nick-gorman:
Two related but separable changes to the new-format templater. The first is the substance of the PR; the second would be better separated but I think the commits are entangled so I'm just leaving combined (sorry).

## 1. Granularity-invariant output set

Previously the new-format templater produced a different *set* of tables depending on regional granularity: the custom-constraint tables were written only at `sub_regions`, since coarser granularities collapse the flow paths they reference. That forced `create_ispypsa_inputs` to track a granularity-dependent output list and left downstream consumers checking for tables that may or may not exist.

This makes the output set identical at every granularity. Where there's nothing to constrain, the custom-constraint tables are output header-only ("all columns, no rows") rather than omitted, so the column shape is invariant and only the row contents differ.

## 2. Wire timeslices templating in

`timeslices.py` and its tests were already on `main` but nothing called it. This adds the call in `create_ispypsa_inputs_template` — `_template_timeslices(load_timeslice_calendar(version), manual_tables["reference_year_sequence"])` — and registers `timeslices` in the output list. Templater tests patch `load_timeslice_calendar` with a two-window stub calendar; the CLI test pins the decoded row count against the shipped 7.5 calendar (content is covered by `test_timeslices.py::test_shipped_calendar_decodes`).

```
src/ispypsa/templater/
├── create_template.py                  → header-only cc tables at coarse granularity;
│                                          wire in timeslices
└── custom_constraints_from_plexos.py   → empty_custom_constraint_tables() + shared column constants
tests/
├── test_cli/test_create_ispypsa_inputs_new_table_formats.py  → verify against list_templater_output_files;
│                                                                 timeslices row-count pin
└── test_templater/test_create_ispypsa_inputs_template.py     → set-equality + header-only checks;
                                                                 timeslice stubs
```

### DIFF
--- a/src/ispypsa/templater/create_template.py
+++ b/src/ispypsa/templater/create_template.py
@@ -252,13 +252,7 @@ def create_ispypsa_inputs_template(
             template["storage_new_entrant"],
             sub_regional_geography,
         )
-        # Custom constraints from PLEXOS are sub-regional export-group limits:
-        # their LHS references sub-region nodes, sub-regional flow paths, and
-        # REZ-located units that only exist as distinct entities at sub_regions
-        # granularity. Once sub-regions are collapsed (nem_regions /
-        # single_region) they have no meaningful representation, so the tables
-        # are emitted header-only ("all columns, no rows") there, keeping the
-        # output set granularity-invariant.
+
         if regional_granularity == "sub_regions":
             template.update(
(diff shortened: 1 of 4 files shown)


EllieKallmier (approved):
Yep all makes sense to me, no notes :)

Issue #123, opened 2026-06-19
Translator timeslice implementation notes (decisions from staging PR #121)


nick-gorman:
## Context

PR #117 (`new-format-translator`) is the omnibus/baseline implementation of the new-format pipeline, including the timeslice work. It is being brought onto `main` as a series of small, separately-reviewable PRs rather than merged wholesale.

PR #121 stages the **templater** half (decoding AEMO's PLEXOS calendar into one window pattern per reference year). While compiling it, we made/validated several decisions that belong to the **translator** half and so land in later staging PRs. This issue records them so they aren't lost between PRs.

Established along the way: within each region and reference year the shipped v7.5 windows form an **exact partition** of the year (every day covered by exactly one timeslice — verified across all 75 region×reference-year groups). `winter_reference` is already the contiguous complement of peak+summer (a fixed `04-01 → 11-01` block for every reference year), with the weather variation carried by the peak/summer windows. The cycle boundary is 1 November; consecutive reference years abut there. The downstream "gap fallback" is therefore defensive code the real data never exercises. The items below lean on that structure holding.

### Enforcement, as of #121

- **Tiling is enforced inline:** `_template_timeslices` calls `_raise_unless_windows_tile_the_year` before returning, so a future calendar that fails to tile fails at template time. The check is boundary-based (sorted by start, each window must end where the next begins, the last wrapping to the first) — no day counting, leap-safe.
- **Two structural assumptions underpin the reference-year attribution**, and are checked by **standalone** guards, *not* inline: `_raise_unless_only_winter_crosses_financial_year` (only winter may span 1 July, so weather-varying timeslices stay within one reference year) and `_raise_unless_winter_is_constant_per_region` (the one boundary-crossing timeslice is identical across reference years, so its attribution can't blur). These run on the real calendar inside `test_shipped_calendar_decodes`, so they're enforced **per workbook version** rather than on every decode — deliberately, to keep the focused decoder tests on minimal data.
- The `timeslices` schema `custom_validation` rules (no-overlap, full-coverage) remain inert until the schema-validation layer is built; the guards above are the active enforcement until then.

## To action in later 117 staging PRs

[...]

PR #126, opened 2026-06-24
Translate the new-format network tables to buses and links

nick-gorman:
This PR does two things: it adds the translator that turns the new-format network tables into PyPSA buses and links, and it uses those tables to pilot a "wildcard" convention for sparse inputs — implemented end-to-end so there is a concrete example to react to before deciding whether to adopt it more broadly.

```
src/ispypsa/
├── translator/
│   ├── network.py            → network tables → buses + links (+ link_timeslice_limits)
│   └── helpers.py            → _resolve_wildcards, the generic wildcard resolver
└── validation/schemas/
    ├── network_transmission_path_limits.yaml           → wildcard keys, nan_fill, coverage rules
    ├── network_expansion_options.yaml                  → wildcard keys, pairing rule
    └── network_transmission_path_expansion_costs.yaml  → wildcard keys, nan_fill, coverage rules
tests/test_translator/
├── test_network.py            → buses/links coverage incl. wildcard and guard cases
└── test_translator_helpers.py → _resolve_wildcards unit coverage
```

## The translation

`network.py` consumes `network_geography`, `network_transmission_paths`, `network_transmission_path_limits`, `network_expansion_options` and `network_transmission_path_expansion_costs`, and produces:

- one bus per modelled geography (REZ buses only when REZs are modelled as discrete nodes),
- one existing link per transmission path,
- expansion links per investment period, from the unified options/costs tables.

Flow paths and REZ connections run through one pipeline — both are just paths.

A link's `p_nom` is the larger of its forward and reverse capacities; every directional, per-demand-condition limit is expressed per unit of that in a `link_timeslice_limits` table (forward limits as `p_max_pu`, reverse as `p_min_pu`), which pypsa_build later expands into per-snapshot series:

```
network_transmission_path_limits:
    path_id  direction  timeslice             capacity
    CQ-NQ    forward    qld_peak_demand       1200
    CQ-NQ    forward    qld_winter_reference  1400
    CQ-NQ    reverse    qld_peak_demand       1440
    CQ-NQ    reverse    qld_winter_reference  1910

links (abridged):
    name            p_nom  p_min_pu  p_max_pu
    CQ-NQ_existing  1910   0.0       1.0       # p_nom = max over both directions

link_timeslice_limits:
    name            attribute  timeslice             value
    CQ-NQ_existing  p_max_pu   qld_peak_demand        0.628   # 1200/1910
    CQ-NQ_existing  p_max_pu   qld_winter_reference   0.733
    CQ-NQ_existing  p_min_pu   qld_peak_demand       -0.754   # -1440/1910
    CQ-NQ_existing  p_min_pu   qld_winter_reference  -1.0
```

Design choices worth knowing before reading the diff:

- `p_nom` as the max of both directions keeps every per-unit limit in [-1, 1], so no timeslice can credit a link above its physical rating.
- The static `p_min_pu`/`p_max_pu` on each link are inert defaults (0.0/1.0); the real limits live entirely in `link_timeslice_limits` under a coverage contract (#123) — every snapshot is covered by a named-timeslice row or a blank-timeslice fallback.
- Expansion links start at `p_nom` 0 and are individually unbounded; the expansion-limit custom constraints (`ispypsa.translator.constraints`) cap the total built across a path's links. Asymmetric options set `p_max_pu`/`p_min_pu` in the option's forward/reverse proportion.

## The wildcard proposal (feedback wanted)

The three sparse tables (limits, expansion options, expansion costs) accept blank key cells as wildcards, resolved most-specific-wins, with blank value cells taking a schema-declared `nan_fill`:

```
network_transmission_path_limits (blank cells are wildcards):
    path_id  direction  timeslice  capacity
    CQ-NQ    forward               1500
    CQ-NQ    reverse               1000
                                   800       # table-wide default

resolved against modelled paths {CQ-NQ, Q1-NQ}:
    path_id  direction  timeslice  capacity
    CQ-NQ    forward               1500      # its own rows win (fewest wildcards)
    CQ-NQ    reverse               1000
    Q1-NQ    forward               800       # inherits the table-wide default
    Q1-NQ    reverse               800
```

`_resolve_wildcards` in `translator/helpers.py` is the generic resolver; the three schema files document the convention (wildcard keys, `*_resolve_unambiguously` and coverage rules, `nan_fill`). The validation rules are declarative-only for now — two runtime guards in `network.py` stand in for the option-pairing rule until schema validation lands.

This is deliberately a reference implementation. The same convention could apply to most sparse ISPyPSA tables (build costs, fuel prices, generator properties); the question for review is whether it earns its keep as a uniform way to express defaults and sparse overrides, or whether it's over-engineering that will be hard to apply consistently. Either way, concrete code seemed a better basis for that call than an abstract proposal.

## Scope note

The translation code is dark — nothing calls it until the orchestrator PR, so there's no change to production behaviour. This PR is stacked on #129 (the `transmission_default_limit` rename) and targets that branch as its base; once #129 merges it retargets to main.

### DIFF
--- a/src/ispypsa/translator/helpers.py
+++ b/src/ispypsa/translator/helpers.py
@@ -1,3 +1,4 @@
+import logging
 import re
 
 import pandas as pd
@@ -154,3 +155,106 @@ def convert_to_numeric_if_possible(df: pd.DataFrame, cols: list[str]) -> pd.Data
         df = df.drop(columns=["temp_col"])
 
     return df
+
+
+def _resolve_wildcards(
+    table: pd.DataFrame,
+    allowed_values: dict[str, list],
+    value_columns: list[str],
+    expected_drops: tuple[str, ...] = (),
+) -> pd.DataFrame:
+    """Expand a sparse "wildcard" table into one row per concrete key combination.
+
+    A key column may be left blank (NaN) to act as a wildcard that applies to
+    every value of that column. ``allowed_values`` lists, for each wildcardable
+    column, the concrete values it may take — the schema's allowed_values /
+    allowed_values_from, resolved to actual values. Each blank cell in those
+    columns is fanned out to every allowed value; a filled cell survives only if
+    it is itself an allowed value, so a row carrying an out-of-set value drops
+    out. Drops are logged at INFO unless the column is named in
+    ``expected_drops`` — for those, dropping is the caller's designed selection
+    (e.g. keeping only investment-period years), not data loss worth surfacing.
+    Key columns absent from ``allowed_values`` (e.g. timeslice) ride along
+    unchanged.
+
+    Once the blanks are filled in, several rows can land on the same key — a
+    specific row and a wildcard one. The row that used the fewest wildcards (the
+    most specific) wins; callers rely on the schema's *_resolve_unambiguously
+    rule to guarantee there is never a tie.
+
+    I/O Example:
+        table (a blank cell is a wildcard):
+            path_id  direction  timeslice  capacity
+            CQ-NQ    forward    peak       1200
+                                           500       # all keys blank: a global default
+            N1-CNSW                                  # blank direction and capacity
+
+        allowed_values = {"path_id": ["CQ-NQ", "N1-CNSW"],
+                          "direction": ["forward", "reverse"]}
+        value_columns = ["capacity"]
+
+        returns:
+            path_id  direction  timeslice  capacity
+            CQ-NQ    forward    peak       1200      # specific row, untouched
+            CQ-NQ    forward               500       # global default fills the gaps...
+            CQ-NQ    reverse               500
+            N1-CNSW  forward               NaN       # ...but N1-CNSW's own blank-capacity
+            N1-CNSW  reverse               NaN       #   row is more specific, so it wins
+    """
+    key_columns = [c for c in table.columns if c not in value_columns]
+    work = table.copy()
+    # Count each row's wildcards now, before the blanks are filled in, so that
+    # after expansion we can keep the row that used the fewest wildcards wherever
+    # rows land on the same key.
+    work["_wildcards"] = sum(work[c].isna().astype(int) for c in allowed_values)
+    # Resolve one wildcardable column per pass; each pass fills that column in.
+    for column, values in allowed_values.items():
+        work = _expand_column(work, column, values, column not in expected_drops)
+    # Most specific (fewest wildcards) wins where rows now share a key.
+    most_specific_first = work.sort_values("_wildcards", kind="stable")
+    resolved = most_specific_first.drop_duplicates(key_columns, keep="first")
+    return resolved.loc[:, list(table.columns)].reset_index(drop=True)
+
+
+def _expand_column(
+    table: pd.DataFrame, column: str, allowed_values: list, log_drops: bool
+) -> pd.DataFrame:
+    """Resolve a single wildcardable column of a wildcard table.
+
+    Splits the rows on whether ``column`` is blank, then recombines: a filled
+    cell is kept only if its value is allowed, while each blank (wildcard) cell is
+    cross-joined with the allowed values so it fans out to one row per value.
+    Splitting first is the trick that keeps any blank out of the merge key (a NaN
+    key would not match itself in a merge). Out-of-set filled cells drop out,
+    logged only when ``log_drops`` is set (see _resolve_wildcards).
+
+    I/O Example:
+        table:                      column = "direction"
+            path_id  direction      allowed_values = ["forward", "reverse"]
+            CQ-NQ    forward
+            N1-CNSW                  # blank: a wildcard
+
+        returns:
+            path_id  direction
+            CQ-NQ    forward         # filled and allowed: kept as-is
+            N1-CNSW  forward         # blank: fanned out to every allowed value
+            N1-CNSW  reverse
+    """
+    is_wildcard = table[column].isna()
+    if log_drops:
+        _log_dropped_values(table.loc[~is_wildcard, column], allowed_values, column)
+    # A filled cell survives only if its value is one of the allowed values...
+    concrete = table[~is_wildcard & table[column].isin(allowed_values)]
+    # ...and a blank cell fans out to one row per allowed value (a cross join).
+    allowed_frame = pd.DataFrame({column: allowed_values})
+    expanded = table[is_wildcard].drop(columns=column).merge(allowed_frame, how="cross")
+    return pd.concat([concrete, expanded], ignore_index=True)
+
+
+def _log_dropped_values(values: pd.Series, allowed_values: list, column: str) -> None:
+    """Log, once, any filled values dropped for falling outside the allowed set."""
+    # tolist() converts numpy scalars to native types so the message reads
+    # "[2025]" rather than "[np.int64(2025)]".
+    dropped = sorted(set(values.dropna().tolist()) - set(allowed_values))
+    if dropped:
+        logging.info(f"Dropped rows whose {column} is not an allowed value: {dropped}")
--- a/src/ispypsa/translator/helpers.py
+++ b/src/ispypsa/translator/helpers.py
@@ -1,4 +1,3 @@
-import logging
 import re
 
 import pandas as pd
@@ -161,7 +160,6 @@ def _resolve_wildcards(
     table: pd.DataFrame,
     allowed_values: dict[str, list],
     value_columns: list[str],
-    expected_drops: tuple[str, ...] = (),
 ) -> pd.DataFrame:
     """Expand a sparse "wildcard" table into one row per concrete key combination.
 
@@ -169,13 +167,12 @@ def _resolve_wildcards(
     every value of that column. ``allowed_values`` lists, for each wildcardable
     column, the concrete values it may take — the schema's allowed_values /
     allowed_values_from, resolved to actual values. Each blank cell in those
-    columns is fanned out to every allowed value; a filled cell survives only if
-    it is itself an allowed value, so a row carrying an out-of-set value drops
-    out. Drops are logged at INFO unless the column is named in
-    ``expected_drops`` — for those, dropping is the caller's designed selection
-    (e.g. keeping only investment-period years), not data loss worth surfacing.
-    Key columns absent from ``allowed_values`` (e.g. timeslice) ride along
-    unchanged.
+    columns is fanned out to every allowed value; a filled cell must itself be
+    an allowed value — anything else raises. Callers filter their designed
+    selections (e.g. costs for disabled elements or non-investment-period
+    years) out before calling, so an out-of-set value reaching this point is
+    bad input data, not selection. Key columns absent from ``allowed_values``
+    (e.g. timeslice) ride along unchanged.
 
     Once the blanks are filled in, several rows can land on the same key — a
     specific row and a wildcard one. The row that used the fewest wildcards (the
@@ -209,7 +206,7 @@ def _resolve_wildcards(
     work["_wildcards"] = sum(work[c].isna().astype(int) for c in allowed_values)
     # Resolve one wildcardable column per pass; each pass fills that column in.
     for column, values in allowed_values.items():
-        work = _expand_column(work, column, values, column not in expected_drops)
+        work = _expand_column(work, column, values)
     # Most specific (fewest wildcards) wins where rows now share a key.
     most_specific_first = work.sort_values("_wildcards", kind="stable")
     resolved = most_specific_first.drop_duplicates(key_columns, keep="first")
@@ -217,16 +214,15 @@ def _resolve_wildcards(
 
 
 def _expand_column(
-    table: pd.DataFrame, column: str, allowed_values: list, log_drops: bool
+    table: pd.DataFrame, column: str, allowed_values: list
 ) -> pd.DataFrame:
     """Resolve a single wildcardable column of a wildcard table.
 
     Splits the rows on whether ``column`` is blank, then recombines: a filled
-    cell is kept only if its value is allowed, while each blank (wildcard) cell is
-    cross-joined with the allowed values so it fans out to one row per value.
-    Splitting first is the trick that keeps any blank out of the merge key (a NaN
-    key would not match itself in a merge). Out-of-set filled cells drop out,
-    logged only when ``log_drops`` is set (see _resolve_wildcards).
+    cell is kept as-is (after checking its value is allowed), while each blank
+    (wildcard) cell is cross-joined with the allowed values so it fans out to
+    one row per value. Splitting first is the trick that keeps any blank out of
+    the merge key (a NaN key would not match itself in a merge).
 
     I/O Example:
         table:                      column = "direction"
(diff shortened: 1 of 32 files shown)


src/ispypsa/translator/helpers.py
```diff
@@ -154,3 +154,116 @@ def convert_to_numeric_if_possible(df: pd.DataFrame, cols: list[str]) -> pd.Data
...
+
+    A key column may be left blank (NaN) to act as a wildcard that applies to
+    every value of that column. ``allowed_values`` lists, for each wildcardable
+    column, the concrete values it may take — the schema's allowed_values /
+    allowed_values_from, resolved to actual values. Each blank cell in those
+    columns is fanned out to every allowed value; a filled cell must itself be
+    an allowed value — anything else raises. Callers filter their designed
+    selections (e.g. costs for disabled elements or non-investment-period
+    years) out before calling, so an out-of-set value reaching this point is
+    bad input data, not selection. Key columns absent from ``allowed_values``
```
EllieKallmier, line 174: Honestly I get what this function is doing but this docstring is confusing to me! Definitely a personal preference thing I think but tbh I'd find a more basic set of input descriptions clearer in this case.

EllieKallmier: (The module docstring in network.py does make this clearer - but as it's a helper I kinda think shouldn't need that?)

src/ispypsa/translator/network.py
```diff
@@ -0,0 +1,734 @@
...
+    limits: pd.DataFrame, all_paths: pd.DataFrame, modelled_paths: pd.DataFrame
+) -> pd.DataFrame:
+    """Drops limit rows for paths configured out of the model.
+
+    REZ-to-parent paths leave the modelled set when REZs are attached to their
+    parent nodes (see _drop_rez_paths_if_not_modelled); their limit rows are
+    config-driven selection, filtered out here (logged at INFO) before wildcard
+    resolution. Only
+    rows naming a known-but-unmodelled path drop: blank (wildcard) path_ids ride
+    through, and a path_id unknown to the paths table survives to
```
EllieKallmier:
```suggestion
    a config-driven selection, filtered out here (logged at INFO) before wildcard
    resolution. Only rows naming a known-but-unmodelled path drop: blank (wildcard) 
    path_ids ride through, and a path_id unknown to the paths table survives to
```

src/ispypsa/translator/network.py
```diff
@@ -0,0 +1,734 @@
...
+    return limits[~dropped]
+
+
+def _resolve_path_limits(
+    limits: pd.DataFrame, paths: pd.DataFrame, default_limit: float
+) -> pd.DataFrame:
+    """Resolves the sparse limits table to one row per modelled path, direction
+    and timeslice, then fills empty capacities with the system default.
+
+    Blank path_id, direction or capacity cells are wildcards (see the
```
EllieKallmier, line 225: Are 'capacity' cells wildcards (terminology)? Just clarifying for myself - I've been thinking of the 'value' columns as kinda different entities from the wildcardable columns. I think prob another case where I'm getting a touch tangled in the docstring just in how some of the terminology gets used - no issue, just my comprehension in the afternoon I reckon

src/ispypsa/translator/network.py
```diff
@@ -0,0 +1,734 @@
...
+def _keep_rows_for_enabled_elements(
+    table: pd.DataFrame, enabled_ids: list[str]
+) -> pd.DataFrame:
+    """Keeps rows whose expansion_id is an enabled element or blank (a wildcard).
+
+    Selecting the enabled elements is config-driven, so it happens before
+    wildcard resolution rather than inside it. Rows for disabled or non-modelled
+    elements drop out here, as do rows for constraint groups (their expansion_ids
+    are constraint_ids, routed to ispypsa.translator.constraints instead). The
+    filtered ids are logged at INFO — a typo'd expansion_id is indistinguishable
```
EllieKallmier, line 489: I think if we assume that inputs have been through validation (including any cross-table validation to cover allowed values/names that are expected to be consistent across inputs?) this is fine and (I think) typos would have been caught and flagged at that point?

src/ispypsa/translator/network.py
```diff
@@ -0,0 +1,734 @@
...
+
+def _check_forward_reverse_share_an_option(options: pd.DataFrame) -> None:
+    """Raises if any element's resolved forward and reverse trace to different
+    options.
+
+    Forward and reverse for an expansion_id must come from one physical option, so
+    the single cost keyed on expansion_id applies to a coherent pair. Wildcard
+    resolution resolves the two directions independently, so a malformed input
+    could pair a forward from one option with a reverse from another; this guard
+    catches that. A blank expansion_option counts as its own label (nunique with
```
EllieKallmier, line 545: this guard catches that after the fact, with the idea being that validation should catch this before the fact right?

src/ispypsa/validation/schemas/network_transmission_path_expansion_costs.yaml
```diff
@@ -20,26 +28,56 @@ description: >
...
     type: string
-    required: true
+    required: false
     allowed_values_from:
       - network_expansion_options: expansion_id
-    description: Identifier for the expandable network element.
+    description: >
+      Identifier for the expandable network element.
+
+      If absent (or empty):
```
EllieKallmier, line 55: Just commenting here cause it's where the thought got jogged but I think this is PR-level question/musing: does the "absent" quality hold the same wildcard-ness as present but "empty"? If a column is allowed to just be totally absent from a table before validation, and the idea is it would just be added as an empty column after, I guess that does imply it's a column of wildcards. I don't know that it's relevant for this table necessarily and will keep thinking about whether there are other tables where this would be problematic (it may not be) but I do think at very least this behavioru should be logged very clearly at validation stage.

src/ispypsa/validation/schemas/network_transmission_path_limits.yaml
```diff
@@ -3,56 +3,82 @@ required: false
...
+  Transmission capacity limits, looked up per path, flow direction and demand
+  condition (timeslice).
 
-  A limit with a timeslice applies to that demand condition only; a limit with
-  no timeslice is a fallback for the conditions no per-timeslice limit covers
-  (see the timeslice column). A path may carry either or both. In the default
-  IASR data, flow-path limits vary by demand condition while REZ-to-subregion
-  connection limits are static, but that reflects the source tables rather than
-  a schema rule.
+  The three key columns — path_id, direction and timeslice — identify the
```
EllieKallmier, line 9: tiny thought - I wonder if defining 'key' as a specific term in the first instance (either by quotation marks or a different word or some other signifier) could be useful? I keep reading it subconsciously as eg 'the three important columns' because of slightly ambiguous definition; I don't think that's an incorrect reading but feels like this is one of those 'it means something kinda specific' moments

src/ispypsa/validation/schemas/network_transmission_path_limits.yaml
```diff
@@ -3,56 +3,82 @@ required: false
...
-      and per-timeslice rows; where a per-timeslice row covers a snapshot its
-      limit takes precedence, and the no-timeslice limit fills the remainder.
+      Wildcard — a fallback across demand conditions, applying at every timeslice
+      no more specific row covers for the same path and direction.
   capacity:
     type: float
     required: false
     units: MW
     gte: 0.0
+    nan_fill: {config: network.transmission_default_limit}
```
EllieKallmier, line 77: sweet I like this

tests/test_translator/test_network.py
```diff
@@ -0,0 +1,1044 @@
...
+    collapsing the link to zero flow in both directions."""
+    ispypsa_tables = _network_tables(csv_str_to_df)
+    ispypsa_tables["network_transmission_paths"] = csv_str_to_df("""
+        path_id,  geo_from,  geo_to,  carrier
+        CQ-NQ,    CQ,        NQ,      AC
+    """)
+    ispypsa_tables["network_transmission_path_limits"] = csv_str_to_df("""
+        path_id,  direction,  timeslice,             capacity
+        CQ-NQ,    forward,    qld_winter_reference,  0
+        CQ-NQ,    reverse,    qld_winter_reference,  1000
```
EllieKallmier, line 310: Piecing things together for my understanding: this input wouldn't normally be allowed (by validation) because not all timeslices are defined (and no empty/wildcard fallback either)... right? So then the link_timeslice_limits would be 'complete' (all timeslices accounted for) because the input is complete (or filled by wildcard). Is that a correct interpretation?

src/ispypsa/translator/network.py
```diff
@@ -0,0 +1,734 @@
...
+
+        existing_links (abridged):
+            isp_name  name              p_nom
+            CQ-NQ     CQ-NQ_existing    1400
+
+        returns:
+            name            attribute  timeslice             value
+            CQ-NQ_existing  p_max_pu   qld_peak_demand       0.857   # 1200/1400
+            CQ-NQ_existing  p_max_pu   qld_winter_reference  1.0     # 1400/1400
+            CQ-NQ_existing  p_min_pu   ,                     -0.714  # fallback, -1000/1400
```
EllieKallmier, line 372: I'm interpretting this empty/nan timeslice in the returned table to imply wildcarding can continue as translator output/model input - is that what you're thinking too? (Or maybe I'm misunderstanding?)

tests/test_translator/test_network.py
```diff
@@ -0,0 +1,1044 @@
...
+    symmetric timeslice = NaN fallback at the default limit, so it tiles the year
+    the same way a data path does (Open-ISP/ISPyPSA#123)."""
+    ispypsa_tables = _network_tables(csv_str_to_df)
+    ispypsa_tables["network_transmission_paths"] = csv_str_to_df("""
+        path_id,  geo_from,  geo_to,  carrier
+        N1-CNSW,  N1,        CNSW,    AC
+    """)
+    ispypsa_tables["network_transmission_path_limits"] = csv_str_to_df("""
+        path_id,  direction,  timeslice,  capacity
+        N1-CNSW,  ,           ,
```
EllieKallmier, line 353: (same just piecing together for myself here) - this 'capacity' value would be set to the default (from config) before being input to translator as a result of validation, so once validation is in place we wouldn't see any NaN/empty 'capacity' values in this table. But at current state makes sense as-is

tests/test_translator/test_network.py
```diff
@@ -0,0 +1,1044 @@
...
+    ispypsa_tables["network_transmission_paths"] = csv_str_to_df("""
+        path_id,  geo_from,  geo_to,  carrier
+        CQ-NQ,    CQ,        NQ,      AC
+        Q1-NQ,    Q1,        NQ,      AC
+    """)
+    ispypsa_tables["network_transmission_path_limits"] = csv_str_to_df("""
+        path_id,  direction,  timeslice,  capacity
+        CQ-NQ,    forward,    ,           1500
+        CQ-NQ,    reverse,    ,           1000
+        ,         ,           ,           800
```
EllieKallmier, line 647: Vague question sitting in the back of my mind: is this blank-row-with-value kinda just a way to get two different default values for this field? (Applied slightly differently ofc). Just wondering about the utility or potential for confusion/double up if so. I guess the distinction in this case is that the config default is the value used for rows with some keys/no values, and this no-row 'capacity' is the value used for no keys/no values (path definitely does exist though). Is that right? 

I don't have any problem with this btw - just I think something to be clear about with this wildcarding and the interactions between any various 'defaults'.

EllieKallmier (approved):
Solid! Have the usual few questions mostly to confirm that I'm understanding flows and the wildcarding process right - I don't think anything blocking from me :)

src/ispypsa/translator/helpers.py
nick-gorman: Yep fair comment.

tests/test_translator/test_network.py
nick-gorman: Yes, thats right in validation would fail on this input, its just kept shorter for testing. But I should add a note to the test, and if there are others like.

src/ispypsa/validation/schemas/network_transmission_path_expansion_costs.yaml
nick-gorman: Good point. As things currently stand yes, an empty column would get inserted as NaN and then act as a wildcard. So if you left out the techology col, it would be like saying these inputs apply equally to all techonologies. I kinda like that but up for debate for sure.

src/ispypsa/translator/network.py
nick-gorman: Yep, maybe not needed but just threw it in for now.

src/ispypsa/translator/network.py
nick-gorman: Yeah, agreed, typos should be gone, here so maybe the logging isn't needed, but I think fine to leave it just in case for now.

src/ispypsa/translator/network.py
nick-gorman: Your right, calling the blank capacity value a wilde card is really wrong. I'll change the wording.

src/ispypsa/translator/network.py
nick-gorman: hmmm, just coming back to this. Maybe better to remove the logging, as it does add noise which overall degrades the loggings usefulness

src/ispypsa/validation/schemas/network_transmission_path_limits.yaml
nick-gorman: Hmmm, I think what's happening here is that we have two types of data in our table category/ID columns which are often used for joining/linking data between tables, kind of like primary or foriegn keys in a relational database, then we just have the values type columns capacity, costs etc. Maybe I've over simplified there, not sure. But anyway we've never adopted any terminolgy for talking about these different types of data, and so claude has just made some up, adopting key in more slang version of how its used relational databases. It made sense to me, but yeah, I see how it might not. I think a better word / defining it explicitly makes sense, but I'm not sure where that would best live.

src/ispypsa/translator/network.py
nick-gorman: Yep, so timeslices are an interesting case where I was thinking to not fully resolve to explicit values in the translator, and allow the model build stage to consume blank values which would act as fallback values if there are gaps in the timeslice definitions. Maybe this gets removed / ends up redundant, if we force timeslice definitions to be complete and cover all model snapshots.

tests/test_translator/test_network.py
nick-gorman: yep

tests/test_translator/test_network.py
nick-gorman: Yes, your understanding is correct. It is a touch confusing though. I think the other thing to note is that the config value is being used for rez transmission  elements where the IASR gives no explicit limit and instead models the limit using a group constraint. So config value should usually be a very high value, but not so high that it causes solve issue. Where as defining a default in the table would more likely be if you were build a toy model and you just wanted to have all transmission elements the same size, its a very niche use case.  Maybe, I just like it cause it shows off the flexibility of the wildcard system, but the actual usefullness might be questionable.

Issue #127, opened 2026-06-25
Managing named new-entrant projects in templater: BOTN - Cethana


EllieKallmier:
The new-entrant set is almost entirely **generic technology archetypes**: `Pumped Hydro (Nhrs storage)`, `CCGT`, `Wind`, etc.  that can be built anywhere. 

**BOTN - Cethana** is the exception: named, site-specific PHES project (fixed 750 MW, 20h/80% design, in TAS) that the IASR workbook gives its own row in _nearly_ every property table, keyed by project name rather than by technology archetype. In most property tables, the values that it is assigned match those given to 'Pumped Hydro (24hrs storage)', which is the 'Technology Type' assigned to BOTN in the summary table; but there are some instances where it doesn't (e.g., build and connection costs, pumping efficiency).

It's also spelled two ways across the v7.5 tables, either 'BOTN - Cethana' or 'BOTN - Cethana - 20h'

### Current handling:

- **New-entrant storage templater** (`src/ispypsa/templater/new_entrants.py`):  BOTN keeps `technology = "Pumped Hydro (24hrs storage)"` (so purpose of 'technology' column - grouping - is preserved). Helper function `_phes_lookup_key` manages the one table merge where this matters, for `pumped_hydro_new_entrant_properties`, where BOTN differs from the generic 24h PHES.
- **Connection costs ** (`src/ispypsa/templater/connection_and_build_costs.py`): BOTN is **dropped** via `_NON_VRE_EXCLUDED_TECHNOLOGIES` in `_canonicalise_non_vre_technologies`. Its connection cost in the forecast is **$0.0**.

#### Thoughts:

- To me makes most sense to explicitly handle BOTN as a special case as needed; just need to decide **how** to do that in each situation
- Dropping, merging and then adding back in BOTN rows? Re-keying temporarily (as done in the new_entrants templater atm)? Overriding 'technology' value for templating process, then manually re-setting after all templating finished? TBD...

EllieKallmier:
New note: #128 (latest update) goes with a manual override of the 'technology' value for BOTN to keep things simple for the moment. 

Todo - update connection costs templating to stop the BOTN explicit drop - it should be able to pass through and be handled as its own technology now with no drama.

PR #128, opened 2026-06-25
Template new entrant property merges

EllieKallmier:
### See below comment for updates - re: BOTN handling

This PR adds merging in of 'property' values for new entrant storage and generation tables. These are the numeric values that map to 'technology'  (see the *_PROPERTY_MAP dicts in mappings for full coverage of what's getting merged in here). Still not outputting the *_new_entrant tables from templater so not 100% wired in yet - coming in next PR(s) :) 

Storage gets split up into pumped hydro and battery chunks to merge in properties whose values are in different tables, then recombines to return a unified 'storage' table still. There's a bit of special handling around BOTN - Cethana, which is the only 'named' new entrant unit and has some bespoke mappings (using 'name' rather than 'technology' to key). Happy to take suggestions on this handling and have opened up #127 issue to explore this in a bit more depth. 

## Where things live

  src/ispypsa/templater/
  ├── new_entrants.py        ← the two orchestrators + the merge machinery + BOTN handling
  ├── mappings.py            ← the property maps (common, battery, PHES)
  ├── create_template.py     ← storage templater now takes the full iasr_tables dict
  └── helpers.py             ← incidental: _fuzzy_map_to_canonical renamed (see below)
  src/ispypsa/iasr_table_caching/local_cache.py   ← register battery_properties + pumped_hydro
  tests/test_workbook_table_cache/7.5/*.csv        ← frozen cache gains the property tables

There's also an incidental rename of `_fuzzy_map_to_canonical` -> `_fuzzy_map_to_allowed_values` because it became more useful to kind of flexibly specify which set of strings to use as the 'choices' than explicitly say one is canon. 

LCF values to be merged in next :)

### DIFF
--- a/src/ispypsa/templater/mappings.py
+++ b/src/ispypsa/templater/mappings.py
@@ -316,18 +316,15 @@ lookup information that can be used to retrieve values.
 
 # - New-format (flag: use_new_table_format=True) per-technology property merge maps -
 
-_GENERATORS_NEW_ENTRANT_PROPERTY_MAP = {
+# Property entries shared by new entrant generators and storage: each looked up by
+# technology from the same IASR table for both.
+_COMMON_NEW_ENTRANT_PROPERTY_MAP = {
     "fom": dict(
         table="fixed_opex_new_entrants",
         technology_col="Technology Type",
         # NOTE: literal double ")" — parsed directly from the v7.5 IASR workbook
         value_col="Base value ($/kW/year))",
         scale=1000.0,
-    ),
-    "vom": dict(
-        table="variable_opex_new_entrants",
-        technology_col="Generator",
-        value_col="Base value",
     ),
     "lifetime_technical": dict(
         table="lead_time_and_project_life",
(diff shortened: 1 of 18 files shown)


EllieKallmier:
Made a change! I don't think this PR had been reviewed yet so to avoid making you review then basically re-review the same stuff I've just bundled an update to how I've handled BOTN - Cethana in here - hope that's not too much stuff in one block.

The main thing to note is just that for now I have jsut set BOTN's 'technology' to 'BOTN - Cethana'; most of the property tables key BOTN's values using that exact string (except for, annoyingly, `pumped_hydro_new_entrant_properties` which uses the long-form 'BOTN - Cethana - 20h'). So there is still one kinda niche declared rename for that table to allow the merges to just run smoothly.

There are some more notes and other thoughts from me in: #127. I don't think this is fully resolved/settled just yet so I'm going to keep that issue open, but  will note that this is the approach I'll take for now (open to alternatives or preferences in a different direction ofc).

src/ispypsa/templater/mappings.py
```diff
@@ -314,6 +314,113 @@
...
+        value_col="Storage capacity (hours)",
+    ),
+    "round_trip_efficiency": dict(
+        table="pumped_hydro_new_entrant_properties",
+        technology_col="Power Station / Technology",
+        value_col="Pumping efficiency (%)",
+    ),
+}
+"""
+New entrant property columns (keys) mapped to the IASR table and columns that contain
```
nick-gorman: I find the docstring after the code a bit confusing. IDK is this a common convention?

src/ispypsa/templater/new_entrants.py
```diff
@@ -58,77 +97,249 @@
...
+    )
+    storage = pd.concat([batteries, phes], ignore_index=True)
+    storage = _merge_properties(storage, iasr_tables, _COMMON_NEW_ENTRANT_PROPERTY_MAP)
+    return storage[_STORAGE_IDENTITY_COLUMNS + _STORAGE_PROPERTY_COLUMNS]
 
 
 # --- shared helpers ---
 
 
+def _merge_properties(
```
nick-gorman, line 218: Having the fuzzy matching inside _merge_technology_keyed_property means its repeated multiple times on the same table, so the logging gets noisy repeatedly reporting the same fuzzy matches. I haven't put too much thought into this, so pitch of salt, but below potential restructure which would mean fuzzy matching and logging runs once per table:

```python

  def _merge_properties(new_entrants, iasr_tables, property_map):                                                                                                                                                                   
      for (table_name, technology_col), props in _group_by_source_key(property_map).items():                                                                                                                                        
          table = iasr_tables[table_name]                                                                                                                                                                                           
          matched_technology = _fuzzy_map_to_allowed_values(                                                                                                                                                                        
              new_entrants["technology"],                                                                                                                                                                                           
              table[technology_col],                                                                                                                                                                                                
              task_desc=f"merging new entrant properties from '{table_name}'",                                                                                                                                                      
          )                                                                                                                                                                                                                         
          for new_col, attrs in props.items():                                                                                                                                                                                      
              _assert_property_table_attrs(table=table, attrs=attrs, property_name=new_col)                                                                                                                                         
              new_entrants = _add_property_value(                                                                                                                                                                                   
                  new_entrants, table, matched_technology, attrs, new_col                                                                                                                                                           
              )                                                                                                                                                                                                                     
      return new_entrants

```

src/ispypsa/templater/new_entrants.py
```diff
@@ -58,77 +97,249 @@
...
+        scale: 1000.0           # $/kW -> $/MW scaling
+
+        returns (new col "fom"):
+            name  technology  fom
+            A     Wind        2000.0
+            B     CCGT        5000.0
+            C     Wind        2000.0
+    """
+    values_by_technology = pd.to_numeric(
+        property_table.set_index(technology_col)[value_col], errors="coerce"
```
nick-gorman: As I understand it no coercing is currently needed. If there was in the next workbook version, say with a typo string in a numeric column, would we prefer the templater to fail loud, rather than silently produce NaNs?

```suggestion
        property_table.set_index(technology_col)[value_col], errors="raise"
```

src/ispypsa/templater/mappings.py
nick-gorman: Oh, I see we have mix of before and after in this file.

nick-gorman (approved):
Looks great Ellie. Very readable and nicely structured. I also agree with the testing mainly at the helper level here, as otherwise it would get quiet verbose.

src/ispypsa/templater/mappings.py
EllieKallmier: yeah this file is a bit of a mess - I might do a restructure to clarify. Earlier versions of this file have had docstrings after but I do prob prefer before and good to just pick one op and stick to it hey :)

src/ispypsa/templater/new_entrants.py
EllieKallmier: Applied this suggestion - slightly modified but using the principle of grouping the properties by the 'source key'. I thought about changing the structure of the mapping dicts to be inherently formatted table-down, but decided against cause I think the current structure is readable, self-documenting and very easy to update :)

PR #130, opened 2026-07-08
Re-sequence new-format timeslice patterns onto snapshots

nick-gorman:
Adds the translator-side timeslice module. It maps each templated timeslice window pattern onto the model's snapshots, producing a `timeslice_snapshots` table (`timeslice_id, investment_periods, snapshots`) that later functionality will use to apply per-timeslice link limits and to scope the temporal custom constraints.

```
src/ispypsa/translator/
└── timeslices.py                        → match snapshots to timeslice windows on month-day
└── timeslices.yaml                      → month-day window semantics; config-to-table validation entry
tests/test_translator/
└── test_timeslice_snapshots.py          → one realistic multi-region/multi-year case plus one test per edge rule
```

### DIFF
--- a/src/ispypsa/translator/timeslices.py
+++ b/src/ispypsa/translator/timeslices.py
@@ -1,0 +1,240 @@
+"""Re-sequences the timeslice patterns and maps them onto model snapshots.
+
+The templated ``timeslices`` table gives one month-day window pattern per
+reference (weather) year. This module assigns a reference year to each model
+financial year using the configured reference_year_cycle — the identical
+assignment the trace pipeline uses, so timeslices stay consistent with the
+demand and VRE traces — expands the assigned patterns into absolute date
+windows, and joins them onto the model's snapshots. The resulting mapping is
+what pypsa_build uses to expand per-timeslice link limits into series and to
+restrict custom constraints to the snapshots their RHS values apply to.
+"""
+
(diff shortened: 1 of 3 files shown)


EllieKallmier (approved):
Yeah this all seems pretty buttoned up to me :) 

A general translator-related thought comes up - that I think we've discussed before but I can't remember the conclusion/if there was one - around validating/checking translator outputs and whether we're going to adopt the same approach here as for the templater outputs. Or even just whether there's value in writing up translator output schemas in a similar way as a mid-term reference or something? 

But no notes on this PR from me, happy to go with this as is :)

nick-gorman:
> Yeah this all seems pretty buttoned up to me :)
> 
> A general translator-related thought comes up - that I think we've discussed before but I can't remember the conclusion/if there was one - around validating/checking translator outputs and whether we're going to adopt the same approach here as for the templater outputs. Or even just whether there's value in writing up translator output schemas in a similar way as a mid-term reference or something?
> 
> But no notes on this PR from me, happy to go with this as is :)

@Ellie, sorry forgot to reply to this. Yep, I think translator output schema make a lot of sense and once we build the infrastructure for running validation on the templater output, re-using it for the translator would be pretty easy. So yeah, I think we could be writing those schema going forward if we wanted to.

PR #132, opened 2026-07-17
Template locational cost factors for new entrants

EllieKallmier:
This PR adds the last two property columns in the new entrants' schemas: lcf_build and lcf_om .

**lcf_build** (build/connection cost factor) is looked up per (geo_id, technology) directly from IASR's precomputed technology_specific_lcfs grid, rather than re-derived from the more granular cost-breakdown-ratio × locational-cost-factor tables. This is a slight simplification (noted in docstring) that doesn't lose any IASR data, but it does mean that if a user hand-edits the raw component tables before templating, those edits won't flow through; only an edit to technology_specific_lcfs itself will. Ofc edits to the templated values themselves will still apply and carry through the rest of the model.  >Am open to changing this to calculate lcf_build from component tables if it feels like a worthwhile addition; however, at the moment we've been working on an assumption that the iasr_tables input to the templater come directly from the (parsed) IASR workbook itself and that we don't expect edits to those inputs, so that's where this simplification comes from.

**lcf_om** is unrelated to that calculation — it's a straight per-zone lookup (O&M costs 3 column of locational_cost_factors), technology-independent. Handled on its own rather than piped into the property-merging pipeline directly because the lookup is geo_id rather than technology: the technology-mapped pipeline applies fuzzy matching that would not be appropriate to apply to geo_id's (which are usually <=4 chars), which should match exactly or not at all. 

src/ispypsa/
├── iasr_table_caching/local_cache.py            ← registers technology_specific_lcfs + locational_cost_factors for the cache builder
├── templater/new_entrants.py                    ← lcf_build / lcf_om merge logic, shared _assert_table_valid primitive
└── validation/schemas/
├── generators_new_entrant.yaml               ← required-tables doc: technology_cost_breakdown_ratios → technology_specific_lcfs
└── storage_new_entrant.yaml                  ← same swap; also drops locational_cost_pumped_hydro_factors (not actually used)

+ the usual tests and test data added :) 

Note: generators_new_entrant/storage_new_entrant still aren't fully wired in as create_template.py outputs yet — they're only consumed internally by connection-cost templating (keeping this PR smaller-ish). That's next to come!


src/ispypsa/templater/new_entrants.py
```diff
@@ -370,6 +404,171 @@ def _set_geo_id(new_entrants: pd.DataFrame) -> pd.DataFrame:
...
+    technology_cols = technology_specific_lcfs.columns.difference(
+        [zone_col, description_col]
+    )
+    long = technology_specific_lcfs.melt(
+        id_vars=zone_col,
+        value_vars=list(technology_cols),
+        var_name="lcf_technology",
+        value_name="lcf_build",
+    ).rename(columns={zone_col: "geo_id"})
+    long["lcf_build"] = pd.to_numeric(long["lcf_build"], errors="coerce")
```
nick-gorman: The below change would bring the error handling into better alignment with _merge_lcf_om, and allow for typos to be explicitly caught rather passed to NaN. 

```suggestion
    long["lcf_build"] = pd.to_numeric(                                                                                                                                                                                                
        long["lcf_build"].replace("Not Applicable", pd.NA), errors="raise"                                                                                                                                                            
    )
```

src/ispypsa/templater/new_entrants.py
```diff
@@ -370,6 +404,171 @@ def _set_geo_id(new_entrants: pd.DataFrame) -> pd.DataFrame:
...
+        new_entrants (BOTN 'technology' already overridden):
+            name                  technology                    geo_id
+            Q1_WH_Far North QLD   Wind                          Q1
+            NQ OCGT Small         OCGT (small GT)               NQ
+            BOTN - Cethana - 20h  BOTN - Cethana                TAS
+
+        technology_specific_lcfs:
+            Cost zone / REZ ID  REZ name / Description  Wind    BOTN - Cethana
+            Q1                  Far North QLD           1.0860  Not Applicable
+            TAS                 Subregional Ref Node    1.0325  100
```
nick-gorman: Just adding in the ocgt to technology_specific_lcfs, is that right?

```suggestion
            Cost zone / REZ ID  REZ name / Description  Wind            OCGT (small GT)  BOTN - Cethana
            Q1                  Far North QLD           1.0860          Not Applicable   Not Applicable
            NQ                  Subregional Ref Node    Not Applicable  1.0801           Not Applicable
            TAS                 Subregional Ref Node    1.0325          Not Applicable   100
```

nick-gorman (approved):
Hi Ellie, 

Looking good to me. Just a couple of minor suggestions from me.

Issue #133, opened 2026-07-23
Non-REZ IDs N0/V0 not templated in network_geography


EllieKallmier:
The NSW and VIC non-REZ IDs N0 and V0 don't appear in the IASR table `renewable_energy_zones`,
so they don't get templated into the `network_geography` table. At the moment this isn't
causing anything to break, and in templating new entrants doesn't cause problems because
we can just use a 'not-subregion' filter that leaves rows with N0/V0 geo_id grouped
with other REZs explicitly defined in the geography table. However, as we add in the
wild-card expansion system and in case other templating/translation steps end up relying
on the `network_geography` table, this might end up causing silent data losses.

dylanjmcconnell:
Hey - not sure if strictly relevant? But just flagging this from the [ADR](https://github.com/Open-ISP/isp-trace-parser/blob/main/docs/adr/001-trace-name-mapping.md) in the trace-parser:

> _AEMO's non-REZ zones (N0, V0) parent at a real subregion like every other zone (AEMO attributes "NSW Non-REZ" to CNSW and "Victoria Non-REZ" to VIC in its new-entrants modelling inputs). This keeps every zone on the same zone -> subregion -> region chain with no special case in the resolver._

EllieKallmier:
Good to note! I think at the moment because they're present in the summary tables we're not losing them entirely, but because we collapse down to just one geo_id we do lose the subregion/region mapping (which is only contained in that network geography table). 

Also though just noticed that in v7.5 and 7.8 workbooks the V0 subregion for new entrants (summary mapping table) is set to 'SEV' - this might be an inconsistency across sources?

Issue #131, opened 2026-07-16
new_entrants_summary v7.5/7.8: NSA pumped-hydro rows have mismatched Regional build cost zone (CSA instead of NSA) [+workbook weirdness]


EllieKallmier:
In the v7.5 IASR workbook's new_entrants_summary table, three pumped-hydro rows with subregion = NSA have their Regional build cost zone column set to CSA instead of NSA. This has passed through to v7.8 workbook as well. 

I'm pretty confident that this is a typo and not intentional: 
- in the 'Regional Build Costs Summary' table PHES rows with subregion=NSA also have NSA listed as the 'Cost zone' value, and
- I haven't been able to find any indication in reports or elsewhere in the workbook that the out-of-place 'CSA' value in the summary table is intentional. 

Current impact:
In the templater for new entrants, I'm treating this as a known typo rather than a real cost-zone split. The templater hardcodes the exception and keys LCF lookups on geo_id for all technologies (in this case, PHES in NSA -> NSA) regardless. PR to come shortly.

I've opened this issue partly to document this decision and partly to remind myself to follow up (see below note).

> **Note:** I haven't yet reported this or followed up with AEMO to confirm whether this is a typo or intentional - we have previously chatted about putting together a list of workbook confusions/bugs/clarifications to put to the ISP team in one go rather than piecemeal, so I think this would go on that list. But not sure if we've started that list yet, so will also note the latest updates in the `isp-workbook-parser` and related documentation of changes from v7.5->7.8 that can be found [here](https://github.com/Open-ISP/isp-workbook-parser/pull/86) and [here.](https://github.com/Open-ISP/project-planning/blob/main/ISP2026/workbook_content_changes_7.8.md)

EllieKallmier:
RELATED (workbook data requiring manual handling): In the `technology_specific_lcfs` table (again, both v7.5 and 7.8), most values are calculated in-place in the workbooks and returned as **factors** but displayed as **percentages** (by Excel cell formatting). So the `isp-workbook-parser` reads those values as factors, i.e., values between 0-1.0 (though sometimes slightly above 1.0 to indicate higher-than-baseline LCF). 

**However:** in classic BOTN special rules, the value for the named 'BOTN - Cethana' new entrant is manually typed in the workbooks as the literal **percentage** value `100.00` (also displayed by Excel as % so in the workbook looks the same as the others). This means though that the workbook parser reads it in as 100.00, so the templater needs to explicitly manage this scale difference.

[...]

PR #134, opened 2026-07-23
Template new entrant regional granularity

EllieKallmier:
Adds functionality to collapse (aggregate) new entrant technologies to the model config's `regional_granularity`.

Changes live here:

```
src/ispypsa/templater/
├── new_entrants.py              ← _collapse_geo_id_to_granularity + orchestrators calling it as their final step
├── create_template.py           ← generators/storage_new_entrant + costs_connection now real template outputs
├── geography.py                 ← _build_geo_region_lookup, deduplicated here
└── transmission.py              ← ...(from here), now imports from geography.py

├── generators_new_entrant.yaml  ← drops fuel_price_mapping (see below) + LCF source table fix
└── storage_new_entrant.yaml     ← same LCF source table fix

src/ispypsa/iasr_table_caching/local_cache.py
                                 ← registers technology_specific_lcfs/locational_cost_factors (missed in prev PR)
```

## Notes

- **The "WOO" duplicate sub-region option is an explicit named case.** Real 7.5 data has one duplicate build option per sub-region for four gas technologies (e.g. "WOO OCGT Small" alongside "CNSW OCGT Small"). Rather than a
  weighted average that implicitly compensates for this, it drops the known rows by name and `_assert_no_unexpected_duplicate_options` raises on any other repeat (unexpected). Documented in comments on Open-ISP/ISPyPSA#131.
- **`fuel_price_mapping` dropped from `generators_new_entrant` entirely**, rather than carried through the collapse. It's not needed until fuel price templating exists - plan is to match fuel prices via `geo_id` +`technology` against a redesigned `costs_fuel_prices` table, now made easier/a more viable option with the planned wild-card schema thing. Two schemas will need matching updates when that happens: `costs_fuel_prices.yaml` and `generators_existing_planned.yaml`.
- **Merged rows are named `"{geo_id} {technology}"`, except BOTN - Cethana**, which keeps its real name.

### Follow-ups for later

- `costs_fuel_prices.yaml` / `generators_existing_planned.yaml` fuel-price-mapping rework (might also rely a bit on #126 and the wildcard handling to properly settle this)
- Pre-existing gap, not fixed here: `N0`/`V0` ("Non-REZ" IDs for NSW/VIC) are used as `geo_id` but never appear in `renewable_energy_zones`, so they're absent from `network_geography` at every granularity — tracked as Open-ISP/ISPyPSA#133, set out as a known exception in the CLI test.

### DIFF
--- a/tests/test_templater/test_create_ispypsa_inputs_template.py
+++ b/tests/test_templater/test_create_ispypsa_inputs_template.py
@@ -62,13 +62,14 @@ def _stub_custom_constraints_tables() -> dict[str, pd.DataFrame]:
 def _new_entrant_property_tables(csv_str_to_df) -> dict[str, pd.DataFrame]:
     """Per-technology property tables the new_entrant templater merges.
 
-    Covers the generator technologies used across the new-format fixtures below
-    (Wind, Large scale Solar PV, OCGT (small GT)) so the property merges resolve.
-    The storage property tables (battery_properties, pumped_hydro_new_entrant_properties)
-    are included too: the fixtures have no storage rows, so the storage subset is empty,
-    but the tables must still be present and non-empty for the merge asserts to pass.
-    Detailed merge behaviour is covered in test_new_entrants.py; here they just
-    need to be present for the wiring to run.
+    Covers every generator/storage technology used across the new-format fixtures
+    below (Wind, Large scale Solar PV, OCGT (small GT), Battery Storage (2hrs
+    storage), Pumped Hydro (24hrs storage)/BOTN - Cethana). technology_specific_lcfs
+    and locational_cost_factors additionally cover every geo_id those fixtures use
+    (Q1, CNSW, SNW, NQ, TAS), including NQ - used by the nem_regions/single_region
+    fixtures to prove a real cross-region merge (see the comments there). Detailed
+    merge behaviour is covered in test_new_entrants.py; here they just need to be
+    present for the wiring to run.
     """
     return {
         "fixed_opex_new_entrants": csv_str_to_df("""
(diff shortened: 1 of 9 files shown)


tests/test_templater/test_create_ispypsa_inputs_template.py
```diff
@@ -402,6 +408,50 @@ def test_create_ispypsa_inputs_template_new_format(csv_str_to_df):
     assert set(costs_connection["geo_id"]) == {"Q1", "CNSW", "SNW"}
     assert len(costs_connection) == 8
 
+    # At sub_regions, geo_id collapse is a no-op: one row per new-entrant summary
```
nick-gorman, line 411: Is this just to add an extra test that regional granularity hasn't mucked up the columns?

nick-gorman (approved):
Looks good Ellie, nothing a mis that I can see.

One small question thats it.

nick-gorman:
Oh, one other comment and question.

Comment: I think the aggregation by averaging is arbitary enough that we could have even left the duplicates in, but also works this way taking them out. Not saying arbitary in a bad way, just that it is.

Question: Did you land on the the WOO just being wollongong, and the duplicates just existing because they added in woollongong locational variat so to speak?

tests/test_templater/test_create_ispypsa_inputs_template.py
EllieKallmier: Yeah kind of, and because this wasn't properly wired in before this point (checking the actual *_new_entrant template outputs) so I was sorta just following the convention that other outputs used here to check the full column set once, and for subsequent granularities just check the row length/a few key features. So I'm not particularly attached to this full column set test and happy to just check shape or something else if this feels clunky/unnecessary to you

EllieKallmier:
> Oh, one other comment and question.
> 
> Comment: I think the aggregation by averaging is arbitary enough that we could have even left the duplicates in, but also works this way taking them out. Not saying arbitary in a bad way, just that it is.
> 
> Question: Did you land on the the WOO just being wollongong, and the duplicates just existing because they added in woollongong locational variat so to speak?

Good point! It is also a bit of extra machinery that yeah is not necessary (and is arbitrary) - maybe in interest of simplifying I'll chop this and just keep doubles.

Re: WOO, essentially yes - I have assumed that it was added on purpose (but not making any assumptions about what the purpose is lol), so it gets kept and templated as any other tech. But it's still flagged in that workbook issue #131  thread and I think worth checking again if/when we do a batch report of those bits, because if it is meant to be a locational variant it's currently not varying at all!

PR #135, opened 2026-08-11
Docstring rework for _resolve_wildcards, distilled into a CLAUDE.md guide

nick-gorman:
Follows up [Ellie's review feedback on #126](https://github.com/Open-ISP/ISPyPSA/pull/126#discussion_r3541517968) that the `_resolve_wildcards` docstring was confusing. This PR rewords that docstring, and since the rewording surfaced lessons that seemed transferable, captures them as a new CLAUDE.md section on documenting generic helpers, then tests the guide by getting claude to apply it again to to a similar function, `_fuzzy_match_names`. Genuinely curious if you think this is a decent improvement across both functions. It's also so pretty verbose so we'd need to watch out it isn't biting too often.

```
CLAUDE.md                            ← new "Docstrings: generic helpers" section
src/ispypsa/translator/helpers.py    ← _resolve_wildcards docstring reworked per the guide
src/ispypsa/templater/helpers.py     ← the guide applied to _fuzzy_match_names as a second worked example
```

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -94,6 +94,79 @@ def _duplicate_for_both_directions(limits: pd.DataFrame) -> pd.DataFrame:
     """
 ```
 
+### Docstrings: generic helpers
+
+Most helpers don't need this section — a one-line summary and an I/O Example
+is right for a helper that names its own columns and serves one caller.
+Apply this guide when a helper is *generic*: the signals are that
+
+- column or key names arrive as parameters instead of appearing literally in
+  the body, so the body alone can't tell the reader what the data means;
+- it has (or is written to invite) several call sites with different domain
+  meanings; and
+- it enforces a rule its callers depend on but don't restate — a precedence
+  order, a validation contract, a guarantee about the output's shape.
+
+The first signal is the gate; the other two raise the stakes. For these
+functions the docstring is the only place the full story exists, so it earns
+a longer, structured treatment — a short narrative, one idea per paragraph,
+in this order:
+
+1. **Summary line: the function's effect, naming the parameters it
+   connects.** "Resolve which row of ``table`` applies to each key
+   combination in ``allowed_values``, returning one explicit row per
+   combination." The summary should earn the function's name — if it is
+   called _resolve_*, the summary should say what gets resolved.
+
+2. **The input's story.** Before any mechanism, explain what the input data
+   is and why it looks the way it does — especially the oddity the function
+   exists to handle ("Rather than write out a row for every key, the user
+   may leave a key cell blank…"). Motivating the weirdness is what makes the
+   mechanism paragraph land.
+
+3. **What each remaining parameter represents**, in domain terms rather than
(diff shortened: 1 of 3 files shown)


src/ispypsa/translator/helpers.py
```diff
@@ -161,23 +161,38 @@ def _resolve_wildcards(
...
-    selections (e.g. costs for disabled elements or non-investment-period
-    years) out before calling, so an out-of-set value reaching this point is
-    bad input data, not selection. Key columns absent from ``allowed_values``
-    (e.g. timeslice) ride along unchanged.
-
-    Once the blanks are filled in, several rows can land on the same key — a
-    specific row and a wildcard one. The row that used the fewest wildcards (the
-    most specific) wins; callers rely on the schema's *_resolve_unambiguously
-    rule to guarantee there is never a tie.
+    """Resolve which row of ``table`` applies to each key combination in
```
EllieKallmier, line 164: I have a few thoughts about this particular docstring, that may or may not be actually useful in a general sense (forewarned):
- It's definitely better described and explains what the function does/doesn't do
- It's still not like super smooth to read out of context - but that is totally fine and I think good, in a way - it's not meant to be read out of context

I do however feel like in this specific case there's just a bit of a disconnect or slight discomfort with phrasing, but that I think could be clarified (for me - this is such a personal observation so hugest grains of salt!!!) by one of the following:
- Formatting the 'derived vocabulary' distinctly, e.g., italics or bold or using single quotes: **key** | *key* | 'key' -> to be aggressively obviously holding a specific meaning in this context
- OR: I was thinking that if ``key_columns`` were a defined input to this function, that would immediately clarify some of the language and what it relates to (to me). With the full understanding that that's prob not the best/cleanest/most efficient (or necessary!!) solution I just wonder if there are some cases where a small refactor to make a function's *function* more obvious could be a win...

Ok big thought dump, again honestly I think this is just a personal taste thing and really not that deep, I generally prefer these versions of the docstring to their old versions and am happy to green-light this general concept of 'give more context when a helper is multi-use' :)

CLAUDE.md
```diff
@@ -94,6 +94,79 @@ def _duplicate_for_both_directions(limits: pd.DataFrame) -> pd.DataFrame:
     """
 ```
 
+### Docstrings: generic helpers
```
EllieKallmier, line 97: As in my other chunky comment, generally I think I am pro- this approach of requiring more in-depth explanation when a helper services multiple call sites/cases. I also though don't think more text is always going to be the best approach for every case (whatever 'best' means lol). 

-> Also I would lean towards adopting an approach to example-writing in docstrings that encourages writing up multiple examples where there are a few distinct edge cases or potentially unclear operations as an accompanying comment :)

EllieKallmier (approved):
Have percolated a bit and left thought comments but in the end I think this is totally reasonable and not a huge deal if not 'perfected' at any point of our process :)

PR #137, opened 2026-08-17
Apply per-timeslice link limits as per-snapshot p_max_pu/p_min_pu series

nick-gorman:
Functionality for adding per timeslice transmission limits to the pypsa network in the pypsa_build module.

Pypsa friendly tables of the form (abreviated data):

```
link_timeslice_limits:
    name            attribute  timeslice        value
    CQ-NQ_existing  p_max_pu   qld_peak_demand  0.857
    CQ-NQ_existing  p_max_pu   ,                1.0     # fallback
    CQ-NQ_existing  p_min_pu   ,                -0.714  # fallback only

timeslice_snapshots:
    timeslice_id     investment_periods  snapshots
    qld_peak_demand  2025                2025-01-13 12:00
```

Get transformed and inserted into the Pypsa model such that:

``` 
network.links_t.p_max_pu:
    investment_periods  snapshots         CQ-NQ_existing
    2025                2025-01-13 12:00  0.857
    2025                2025-01-15 12:00  1.0
        
network.links_t.p_min_pu:
    investment_periods  snapshots         CQ-NQ_existing
    2025                2025-01-13 12:00  -0.714
    2025                2025-01-15 12:00  -0.714
```

A few points:

- The implemention is defensive such that if an non-extendable link doesn't have a limit value for each snapshot, from an explicitly named timeslice or a NaN fallback, then a error is raised. Might not be needed when validation is fully in-place, but it was an easy way to satisfy myself about the correctedness for now.
-  Also reached into the translator and changed how zero-capacity new corridors are treated, now they aren't a special case and just get timeslice limits set to zero.
- One known gap: a limits table whose `timeslice` column is blank in every row comes back from CSV as float64 and can't be merged onto an empty `timeslice_snapshots`. That's a general CSV-round-trip typing problem (#138) and the fix belongs at the read boundary; an `xfail` test records it here.
- I also realised that I was using `timeslice` and `timeslice_id` as column names, this PR works around it, but #140 will clean up to just use `timeslice` everywhere.
```
src/ispypsa/pypsa_build/links.py                          → new optional args + the per-snapshot expansion
src/ispypsa/translator/network.py                         → zero-capacity corridors get per-unit 0 instead of being skipped
tests/test_model/test_add_links_with_timeslice_limits.py  → new
tests/test_translator/test_network.py                     → zero-capacity test now expects the 0 rows, in the templater's named-timeslice shape
```

### DIFF
--- a/src/ispypsa/pypsa_build/links.py
+++ b/src/ispypsa/pypsa_build/links.py
@@ -250,12 +250,16 @@ def _raise_if_snapshots_uncovered(limits_per_snapshot: pd.DataFrame) -> None:
     if uncovered.empty:
         return
     pairs = sorted(set(zip(uncovered["name"], uncovered["attribute"])))
-    first = uncovered.loc[:, ["investment_periods", "snapshots"]].head(5)
+    first = uncovered.head(5)
+    first_snapshots = [
+        f"({period}, {snapshot})"
+        for period, snapshot in zip(first["investment_periods"], first["snapshots"])
+    ]
     raise ValueError(
         f"link_timeslice_limits leaves {len(uncovered)} (link, attribute, snapshot) "
         f"combination(s) undefined: no fallback (blank-timeslice) row and no named "
         f"timeslice active there. Affected (link, attribute): {pairs}. "
-        f"First uncovered snapshots:\n{first.to_string(index=False)}"
+        f"First uncovered (investment_period, snapshot): {', '.join(first_snapshots)}"
     )
 
 
--- a/src/ispypsa/pypsa_build/links.py
+++ b/src/ispypsa/pypsa_build/links.py
@@ -21,20 +21,32 @@ def _add_links_to_network(
     make link_timeslice_limits and timeslice_snapshots required and drop the
     None handling in _build_link_pu_overrides.
 
-    Args:
-        network: The `pypsa.Network` object
-        links: `pd.DataFrame` with `PyPSA` style `Link` attributes.
-        link_timeslice_limits: `pd.DataFrame` with per-timeslice per-unit
-            limits (columns name, attribute, timeslice, value). Required on
-            the new-format path, where the links table's p_max_pu / p_min_pu
-            are placeholders it overrides at every snapshot; omitted on the
-            old-format path.
-        timeslice_snapshots: `pd.DataFrame` mapping timeslice_ids to the
-            snapshots they are active at (columns timeslice_id,
-            investment_periods, snapshots). Required whenever
-            link_timeslice_limits is given.
+    I/O Example (new-format path):
+        links (the p_max_pu / p_min_pu here are placeholders):
+            name            bus0  bus1  carrier  p_nom  p_max_pu  p_min_pu  p_nom_extendable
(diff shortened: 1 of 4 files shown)


src/ispypsa/translator/network.py
```diff
@@ -370,21 +368,24 @@ def _translate_timeslice_limits_to_pu(
             CQ-NQ_existing  p_max_pu   qld_peak_demand       0.857   # 1200/1400
             CQ-NQ_existing  p_max_pu   qld_winter_reference  1.0     # 1400/1400
             CQ-NQ_existing  p_min_pu   ,                     -0.714  # fallback, -1000/1400
+
```
EllieKallmier, line 371: Could potentially benefit from having a zero-p_nom link in the I/O Example for full illustration?

src/ispypsa/pypsa_build/links.py
```diff
@@ -1,16 +1,362 @@
...
+
+
+def _split_pairs_by_uncovered_extent(
+    limits_per_snapshot: pd.DataFrame, uncovered: pd.DataFrame
+) -> tuple[list[tuple[str, str]], list[tuple[str, str]]]:
+    """Splits the (link, attribute) pairs with uncovered snapshots into those
+    uncovered at every snapshot and those uncovered at only some.
+
+    I/O Example:
+        limits_per_snapshot: A p_max_pu at 2 snapshots (both NaN),
```
EllieKallmier, line 296: I get this and am not against summarised I/O examples when it's this many layers deep + likely to be repetitive of caller's example, but in this case I think the ('name', 'attribute') structure of the returned tuples is a little bit lost. Just cause 'A p_max_pu at 2 snapshots' could just as easily read as a sentence as a 'name, attribute at 2 snapshots' description.
```suggestion
        limits_per_snapshot (abbr.): 
            name    attribute   ... value
            A       p_max_pu    ...         # uncovered
            A       p_min_pu    ...         # uncovered
            A       p_min_pu    ... 0.9
            A       p_min_pu    ...         # uncovered
        
        uncovered: DataFrame with tagged rows (above)
```
(or something) to illustrate directly the return structure maybe? Not hugely important though tbh

tests/test_model/test_add_links_with_timeslice_limits.py
```diff
@@ -0,0 +1,378 @@
...
+
+    _add_links_to_network(network, links, link_timeslice_limits, timeslice_snapshots)
+
+    expected = csv_str_to_df("""
+        Link,            p_max_pu,  p_min_pu
+        CQ-NQ_existing,  1.0,       0.0
+        CQ-NQ_option_1,  0.9,       -0.6
+    """).set_index("Link")
+    # The existing link's static values are the untouched placeholders; its
+    # real limits live in links_t (covered by the tests above).
```
EllieKallmier, line 301: Noting out loud for myself so I don't forget - PyPSA always chooses the time-varying (series) version of an attribute when both series and static values are defined for that attribute :)

EllieKallmier (approved):
nice, all makes sense and looks good to me :) comments are vibe-based and non blocking haha

Issue #142, opened 2026-08-25
Minimum stable level values for coal generators


EllieKallmier:
In version 7.5+ of the IASR workbook, the ``coal_minimum_stable_level`` table has three candidate columns that we could use to set the 'default' templated `minimum_load` value:
- `_IASR 2023 (Backcasting)`
- `_Typical Lowest Band`
- `_Minimum Continuous Operating Level`

For the moment: I set the property merge to use  the `_Typical Lowest Band` column, based on the below copied IASR content. Felt like a bigger call to talk about implementing custom constraints/formulations for using multiple values here - and relates a bit to our discussion the other day about the time-sequential vs capacity expansion detail and complexity - so just went with the static opinionated choice for now.

From the [IASR (2025)](https://www.aemo.com.au/-/media/files/stakeholder_consultation/consultations/nem-consultations/2024/2025-iasr-scenarios/final-docs/2025-inputs-assumptions-and-scenarios-report.pdf?rev=63268acd3f044adb9f5f3a32b6880c27&sc_lang=en) pages 121-122 (emphasis added by me):

> **Additional coal generator minimum stable levels**
> Two new series of coal generator minimum stable levels have been provided in the 2025 Inputs and Assumptions
> Workbook:
> - Typical Lowest Band – this value is representative of the power level typically offered in the lowest band of a generator’s bid stack. This represents a ‘preferred’ minimum power level, where negative price exposure risk is not material. Across the coal fleet, this is a reduction of 527 MW (6%) in baseline minimum stable levels compared to the previous IASR and is a reflection of growing negative price exposure risk.
> - Minimum Continuous Operating Level – this forward-looking value is representative of the minimum continuous power level which the generator could foreseeably offer in their bid stack in response to increasing negative price exposure risk. This represents a ‘tested’ bid at this power level which was dispatched for sustained periods as part of typical generation (not during startup or shutdown) for the specific unit, or it has been observed that most units at the station have demonstrated that capability. Across the coal fleet, the observed potential reduction in minimum stable levels from the current baseline is a 1,317 MW (20.7%) reduction relative to the previous IASR.
> 
> The typical lowest band for existing generators is based on AEMO’s analysis of historical generation and operational experience and **will be applied in the time-sequential model** as a representation of a generator’s typical minimum stable level. The minimum continuous operating level values **may be applied in the time-sequential model** to represent the **future behaviour** of coal generators, which is expected to become more likely as coal generators increasingly seek to avoid exposure to negative prices.

PR #143, opened 2026-08-27
Template existing and planned (ECAA) generators

EllieKallmier:
Adds `generators_existing_planned` to the new-format templater, and lays the
generator/storage split that `storage_existing_planned` will build on. A bit chunkier
than ideal prob but should be pretty similar to the new entrants so (hope) not too
much review work...

Everything comes from one IASR table —
`existing_committed_anticipated_additional_generator_summary` — which already lists one
row per real generating unit (DUID-level), generators and storage together. So the module
splits that table in two, then merges per-unit properties onto the generator side from six
further IASR tables. `storage_existing_planned` is left at just the split and column rename
here; property merges for that one coming up.

```
src/ispypsa/templater/
├── existing_planned.py      → new module: gen/storage split, PHES routing, property merges
├── mappings.py              → _GENERATORS_EXISTING_PLANNED_PROPERTY_MAP; key_col rename
├── helpers.py               → shared helpers consolidated here (see Refactors)
├── create_template.py       → wires the new table into the new-format branch
└── geography.py, transmission.py, new_entrants.py
                             → helpers moved out to helpers.py
src/ispypsa/iasr_table_caching/local_cache.py
                             → registers the seven newly-consumed IASR tables
tests/test_workbook_table_cache/7.5/
                             → frozen cache rebuilt with those tables
```

## Things worth knowing before reading the diff

**PHES routing can't rely on `Technology Type` alone.** Wivenhoe and Shoalhaven are
labelled plain "Hydro" in the summary, but have real entries in the PHES properties table
(and the numbers cross-check — Wivenhoe's 3000 MWh / 285 MW ≈ the properties table's 10h).
Routing by technology string alone would send them to generators and silently never read
their PHES properties. So the split is a technology match *union* name-presence in the PHES
properties table -> noting it could just be one or the other

`_validate_phes_routing` then fails loud if any PHES station is missing from the summary
altogether — the only way to catch a future Wivenhoe-shaped surprise, since routing's own
presence check can't distinguish "absent" from "missed". `Lower Tumut` is the one
documented tolerated exception: it's Tumut 3's reversible pump-turbine subset (units 1-3),
see [#131](https://github.com/Open-ISP/ISPyPSA/issues/131) comment thread.

**Two things to note**:

- `minimum_load` for coal now comes from **Typical Lowest Band**. See [#142](https://github.com/Open-ISP/ISPyPSA/issues/142)
- **`KiataWF1` vs `KIATAWF1`** is handled as a named-constant correction, not by fuzzy
matching — `fuzz.ratio` scores it 50, well below any threshold that would be safe to apply
across 642 IDs. It's the only ID mismatch in the whole set. More detail in comments.

**Known gap - not fixed here** 25 REZ-located rows use sub-zone ids (`Q8a`/`Q8b`/
`Q8c`) that `renewable_energy_zones` doesn't carry — only the parent `Q8`. Same class as
the existing `#133` non-REZ placeholder gap, different cause, so it follows the same
precedent: named as an exception in the CLI-level geo_id coverage test rather than solved
here. Fix to come shortly - also flagging a future need to reassess when we move to v7.8
workbook which introduces a (differently-purposed/different coverage and impact) split up of
`N9` -> `N9a` / `N9b`.

## Refactors

Mostly consolidation, so both new-format modules draw on one set of helpers instead of
reaching sideways into each other:

- Moved into `helpers.py`:
  - `_set_geo_id` (was duplicated in both modules),
  - the geo_id/granularity trio from `geography.py`,
  - `_assert_table_valid` and the property-map machinery from `new_entrants.py`, and
  - `_build_geo_region_lookup` from `transmission.py`
- `_map_geo_id_to_granularity` now handles all three granularities and leaves REZ geo_ids
  untouched in every case, so that rule lives in one place rather than being reimplemented
  per module.
- New entrant's property maps rename `technology_col` → `key_col`, so one merge helper
  serves both modules. The maps gain a `numeric` flag (default true) controlling whether a
  merged value is coerced with `pd.to_numeric(errors="raise")` — an input-side typo guard;
  only `commissioning_date` opts out.
  - There's a note-to-self in there that I'm happy to take opinions on re: float vs int
  dtype enforcement (whether to implement here or let validator apply) - my thought is
  towards implementing here to keep validator's scope tight but it's a small fix thing
  so just left simple as-is for the moment.
- `commissioning_date` gets an explicit reformat: the parser emits ISO strings, the schema
  declares `%d/%m/%Y`.


src/ispypsa/templater/existing_planned.py
```diff
@@ -0,0 +1,498 @@
...
+# Tumut 3 has a real pump/non-pump unit split that the summary and phes_properties
+# tables can't currently be reconciled on by name. _validate_phes_routing tolerates
+# this as known unmatched (rather than renaming units to assign as PHES) as part of
+# an interim simplification to treat all Tumut 3 units as generators.
+# See Open-ISP/ISPyPSA#131 comment thread.
+_KNOWN_UNMATCHED_PHES_STATIONS = {"Lower Tumut"}
+
+# Case mismatch between maximum_capacity's IASR ID and the summary's: the 'safe'
+# fuzzy-matching threshold (90) would miss (fuzz.ratio("KiataWF1", "KIATAWF1") == 50).
+_MAXIMUM_CAPACITY_ID_TYPO_FIX = dict(
```
nick-gorman, line 113: I think this constant would read better with a generic name like IASR_TYPO_FIXES, because then the fixes themselves are defined on a table by table basis. Does that make sense?

src/ispypsa/templater/existing_planned.py
```diff
@@ -0,0 +1,498 @@
...
+            IASR ID / DLT names  Power Station  Technology Type      REZ ID  Sub-region
+            BW01                 Bayswater      Steam Sub Critical   NA      CNSW
+
+        maximum_capacity_existing_committed_anticipated_additional_generators:
+            IASR ID  Installed capacity (MW)  Commissioning date
+            BW01     660.0                    NaN
+
+        returns:
+            name  power_station  geo_id  capacity  commissioning_date
+            BW01  Bayswater      CNSW    660.0     NaN
```
nick-gorman, line 151: Maybe slightly reformatted and expanded I/O example like below?

```suggestion
      I/O Example (subset of columns):                                                                                                                                                                                              
          iasr_tables:                                                                                                                                                                                                              
              existing_committed_anticipated_additional_generator_summary:                                                                                                                                                          
                  IASR ID / DLT names  Power Station  Technology Type     REZ ID  Sub-region                                                                                                                                        
                  BW01                 Bayswater      Steam Sub Critical  NA      CNSW                                                                                                                                              
                                                                                                                                                                                                                                    
              maximum_capacity_existing_committed_anticipated_additional_generators:                                                                                                                                                
                  IASR ID  Installed capacity (MW)  Commissioning date                                                                                                                                                              
                  BW01     660.0                    NaN                                                                                                                                                                             
                                                                                                                                                                                                                                    
              ... plus the other property tables (see _GENERATORS_EXISTING_PLANNED_PROPERTY_MAP)                                                                                                                                    
                                                                                                                                                                                                                                    
          regional_granularity: "nem_regions"                                                                                                                                                                                       
                                                                                                                                                                                                                                    
          sub_regional_geography:                                                                                                                                                                                                   
              geo_id  geo_type   region_id                                                                                                                                                                                          
              CNSW    subregion  NSW                                                                                                                                                                                                
                                                                                                                                                                                                                                    
          returns:                                                                                                                                                                                                                  
              name  power_station  geo_id  capacity  commissioning_date                                                                                                                                                             
              BW01  Bayswater      NSW     660.0     NaN    # CNSW -> NSW via sub_regional_geography
```

src/ispypsa/templater/existing_planned.py
```diff
@@ -0,0 +1,498 @@
...
+    generators: pd.DataFrame,
+    iasr_tables: dict[str, pd.DataFrame],
+    property_spec: dict,
+) -> pd.DataFrame:
+    """Assigns minimum_load from one technology-specific table, bounded to its own technologies.
+
+    Only generators whose ``technology`` appears in the table's own 'Technology Type'
+    column are fuzzy-matched and
+
+    I/O Example (property_spec = _COAL_MINIMUM_LOAD_PROPERTY; value_col abbreviated,
```
nick-gorman, line 463: I think I have a preference for expanding the I/O examples. So in this case showing property_spec and maybe nesting coal_minimum_stable_level under an iasr_tables: headning. Up to you if you think its worth it.

src/ispypsa/templater/existing_planned.py
```diff
@@ -0,0 +1,498 @@
...
+    I/O Example:
+        names:      pd.Series(["HUNTER1", "BW01", "SOMESTATION1"])
+        table_keys: pd.Series(["SOMESTATIONl", "BW01", "HUNTER1"])
+        table_name: "heat_rates_..."
+
+        returns:    pd.Series(["HUNTER1", "BW01", "SOMESTATIONl"])
+    """
+    resolved = _fuzzy_match_names(
+        names,
+        table_keys,
```
nick-gorman, line 388: Maybe this is over kill. But the fuzzy matching would be slightly hardended by filtering out known storage unit names from table_keys before passing to _fuzzy_match_names.

src/ispypsa/templater/existing_planned.py
```diff
@@ -0,0 +1,498 @@
...
+            Borumba         # matches after _BORUMBA_FULL_NAME_MAP
+            Lower Tumut     # excused by _KNOWN_UNMATCHED_PHES_STATIONS
+
+        -> no error
+
+        A "Some New Station" row in phes_properties instead raises:
+            ValueError: PHES properties station(s) not found in the summary:
+                        ['Some New Station']
+    """
+    if summary.empty:
```
nick-gorman, line 276: Not a big one, but I reckon, given that this is the templater we could assume summary always contains some data. I think if you did change this is would just impact a couple of tests.

src/ispypsa/templater/existing_planned.py
```diff
@@ -0,0 +1,498 @@
...
+            IASR ID  Heat rate (GJ/MWh)
+            BW01     10.05
+            HUNTER1  10.93
+
+        returns (one new column per property_map key):
+            name     power_station         capacity  commissioning_date  heat_rate
+            BW01     Bayswater             660.0     NaN                 10.05
+            HUNTER1  Hunter Power Station  375.0     2025-08-01          10.93
+    """
+    if generators.empty:
```
nick-gorman, line 339: Also here?

src/ispypsa/templater/existing_planned.py
```diff
@@ -0,0 +1,498 @@
...
+            IASR ID  Technology Type       Min Stable Level (MW)
+            ANGAS1   Reciprocating engine  3.0
+
+        returns (adds minimum_load):
+            name    technology            minimum_load
+            BW01    Steam Sub Critical    260.0
+            ANGAS1  Reciprocating engine  3.0
+            Q1G1    Large scale Solar PV  NaN   # neither coal nor gas
+    """
+    if generators.empty:
```
nick-gorman, line 439: Also here?

src/ispypsa/templater/mappings.py
```diff
@@ -709,58 +709,106 @@
...
+    ),
+    "heat_rate": dict(
+        table="heat_rates_existing_committed_anticipated_additional_generators",
+        key_col="IASR ID",
+        value_col="Heat rate (GJ/MWh)",
+    ),
+    "closure_year": dict(
+        table="expected_closure_years",
+        key_col="IASR ID",
+        value_col="Expected Closure Year (Calendar year)",
```
nick-gorman, line 812: Could replace numeric with a declared dtype so that the closure can be output as a int rather than a float. Maybe a nitpick, hmmm.

```suggestion
        value_col="Expected Closure Year (Calendar year)",
        dtype="int"
```

src/ispypsa/templater/existing_planned.py
```diff
@@ -0,0 +1,498 @@
...
+        Q8 Battery - 2h  Q8 Battery     Battery Storage (2hrs storage)
+
+Building generators_existing_planned:
+    1. Splits the summary's rows into generators and storage — see
+       _is_existing_planned_storage_row.
+    2. Renames the carried-over spine/identity columns to their schema names, derives
+       geo_id (REZ ID with Sub-region fallback — see helpers._set_geo_id) and relabels
+       it to ``regional_granularity`` (REZ-located rows stay untouched at every
+       granularity, matching new_entrants.py's convention — see
+       _relabel_geo_id_to_granularity).
```
nick-gorman, line 30: Not sure _relabel_geo_id_to_granularity is used anywhere.

src/ispypsa/templater/existing_planned.py
```diff
@@ -0,0 +1,498 @@
...
+import pandas as pd
+
+from ispypsa.templater.helpers import (
+    _apply_known_value_replacement,
+    _assert_table_valid,
+    _fuzzy_match_names,
+    _get_property_value_map,
+    _group_properties_by_source,
+    _is_storage_row,
+    _is_subregion_geo_id,
```
nick-gorman, line 51: _is_subregion_geo_id unused?

nick-gorman (approved):
@EllieKallmier, generally I think this looks great and was pretty easy to follow. I just found a few pretty small things, I'll leave it up to you what you reckon is worth actioning.

src/ispypsa/templater/existing_planned.py
EllieKallmier: Yep I agree! I'll leave this for now as I have an update coming up that this name change will fit very nicely into :)

src/ispypsa/templater/existing_planned.py
EllieKallmier: I *might* leave this one for the moment - but definitely open to remove in future - just for a little (low cost?) hedge. Because this is catching only when the 'generators' part of the summary table is empty rather than the whole summary table it's just a tiny bit less of a given(?). Though I totally accept it's defensive and we're not at the stage of pre-filtering inputs or anything so if you do feel like the cons outweigh the (admittedly minute) pros I will happily delete hahaha 😆

src/ispypsa/templater/mappings.py
EllieKallmier: I have also been wondering about this - at the moment I'm not 100% convinced that the benefits of doing something like this (non-float looking year values is kinda the main one) outweighs the added complexity due to NaN presence. I'm also thinking about the round-trip when tables get saved to and read from csv (with potentially still empty columns) and needing to keep re-setting dtypes at those boundaries, + validation probably will (/could) have some type-setting role I imagine. 

There's also a kind of question about where this mapping 'faces'/what it's kind of telling you about, which at the moment has been pretty much all source-related. A new field that's telling you what will happen to the output is a slightly different direction - but that's a very very small difference (and could be argued that it's still telling about the source too) imo.

Anyway I might open an issue just to keep this comment + thoughts somewhere and am totally open to implementing this if we decide it's worth it!

PR #144, opened 2026-09-03
Fix new entrant renaming

EllieKallmier:
Reworked the new entrants' renaming after regional granularity collapse/aggregation.

Previously: was just setting the new names to a standard '{geo_id} {technology}' formula; this lost some detail for units whose pre-collapse naming convention had different structures for representing technologies. Wasn't causing bugs or issues at this stage - but for mapping against constraints from plexos files or other AEMO-sourced name-keyed values, losing the names' structure had potential to be annoying AND was an easy fix. (+ keeps consistent with non-renamed units e.g. those in REZs).

Now: the actual 'old' geo_id string in a unit name gets re-keyed to the new geo_id (collapsed), keeping the rest of the name as-is. The first 'name' value for each collapsed-group is kept upon aggregation; this is the name that gets used to re-key. This implementation does assume that the naming convention for aggregated units (grouped on technology, geo_id and optionally resource_type) is the same for all units in each group, and that the 'new' geo_id is correctly assigned regardless of the old geo_id-like string prefix in each name. Based on current 7.5 workbook data this holds. Special case 'BOTN - Cethana' doesn't get aggregated with anything else so name gets passed through unchanged. 

Example to illustrate the difference:

| Original name | Technology | New geo_id | Prev version 'name' | New version 'name' |
|---|---|---|---|---|
| CSA Biomass | Biomass | SA | SA Biomass | SA Biomass |
| SNW OCGT Large | OCGT (large GT) | NSW | NSW OCGT (large GT) | NSW OCGT Large |
| SQ Battery - 2h | Battery Storage (2hrs storage) | QLD | QLD Battery Storage (2hrs storage) | QLD Battery - 2h |

Built off #143


dylanjmcconnell:
Hey Ellie - 

As discussed - perhaps could expand this out / make it slightly more readable?

```python
    return new_entrants.groupby(
        group_key_columns + ["geo_id"], dropna=False, as_index=False
    ).agg({"name": "first", **{col: "mean" for col in value_columns}})
```

Something like:

```python
name_aggregation_rule = {"name": "first"} 
value_column_aggregation_rules = {col: "mean" for col in value_columns}
return new_entrants.groupby(
          group_key_columns + ["geo_id"], dropna=False, as_index=False
      ).agg(name_aggregation_rule | value_aggregation_rules)
```

Or variation of ..  

Also maybe move :

`set(old_geo_ids) | _EXTRA_SUBREGION_IN_NAMES`

From `collapse_geo_id_to_granularity`  to `_rekey_names_to_collapsed_geo_id` 

Otherwise looks good!

